PR #5, opened 2020-02-10
add alter after

vaibhavmule:


src/masonite/orm/grammar/BaseGrammar.py
```diff
@@ -93,6 +94,7 @@ def _compile_alter(self):
                 if column.length
                 else "",
                 nullable="" if column.is_null else " NOT NULL",
+                after=" AFTER " + "`{0}`".format(column.after_column) if column.after_column else ""
```
vaibhavmule: @josephmancuso what could be a better way to write it?

josephmancuso: you have to put this string inside an `after_column_string()` method on the `MySQLGrammer` class and then change this line to 

```
after=self.after_column_string().format(after=column.after_column) if column.after_column else ""
```

On the `MySQLGrammer` class create a method like this:

```python
def after_column_string(self):
    return "AFTER {after}"
```

josephmancuso (changes requested):
Awesome. Just a small change to make the code more adaptable

josephmancuso:
**Also whenever you create a method on the MySQLGrammer class, be sure to also put the same exact method on the `GrammerExample` class so we know what methods to add to the other grammars**

vaibhavmule:
@josephmancuso where is `GrammarExample`?

src/masonite/orm/grammar/BaseGrammar.py
```diff
@@ -93,6 +94,8 @@ def _compile_alter(self):
                 if column.length
                 else "",
                 nullable="" if column.is_null else " NOT NULL",
+                after=self.after_column_string().format(after=column.after_column) if column.after_column else ""
```
josephmancuso: actually just modify this `after=self._compile_column(column.after_column)`

src/masonite/orm/grammar/GrammarExample.py
```diff
@@ -84,3 +84,6 @@ def column_string(self):
 
     def value_string(self):
         return "'{value}'{seperator}"
+
+    def after_column_string(self):
+        return " AFTER `{after}`"
```
josephmancuso, line 89: With the change above you can remove the back ticks here

josephmancuso (changes requested):

josephmancuso:
the reason we make this change is it will be easier to manage new grammars and also the column formatting can change. MySQL is 
```
`column`
```

but mssql is 

```
[column]
```

src/masonite/orm/grammar/BaseGrammar.py
vaibhavmule: That will just have ``name``

PR #36, opened 2020-02-29
Adding timestamps fields in Blueprint statement. closes #14

### LINKED ISSUE #14: Schema: Add a basic created_at and updated_at column with a timestamps schema column

Marlysson:
Some steps are done to declare timestamp fields using Blueprint statement.

- [x] Create a basic declaration on create table ( without: default values, time zones )
- [x] **.use_current()** method to define whether call "CURRENT_TIMESTAMP" statement in query. 
- [x] Decide if **default()** method will be used. Setting to "CURRENT_TIMESTAMP".
- [x] Add the **nullable()** method in **timestamp** column type.

To update the **updated_at** field in MySQL there's the "ON UPDATE" to update the field by a trigger. Will this done by an update in python code or SQL statement?

### DIFF
--- a/src/masonite/orm/blueprint/Blueprint.py
+++ b/src/masonite/orm/blueprint/Blueprint.py
@@ -18,6 +18,8 @@ class Column:
         self.after_column = None
         self.old_column = ""
         self._action = action
+        self.current_timestamp = False
+        self.default_value = None
 
     def nullable(self):
         self.is_null = True
@@ -40,6 +42,13 @@ class Column:
         self.after_column = after_column
         return self
 
+    def use_current(self):
+        self.use_current = True
+        return self
+
+    def default(self, value=None):
+        if value:
+            self.default_value = value
 
 class Blueprint:
     def __init__(self, grammar, table="", action=None, default_string_length=None):
@@ -111,6 +120,14 @@ class Blueprint:
     def datetime(self, column, nullable=False):
         self._last_column = self.new_column("datetime", column, None, nullable)
         self._columns += (self._last_column,)
+        return self
+
+    def timestamps(self):
+        created_at = self.new_column('timestamp', 'created_at', None, nullable=False).use_current()
+        updated_at = self.new_column('timestamp', 'updated_at', None, nullable=False).use_current()
+
+        self._last_column = updated_at
+        self._columns += (created_at, updated_at,)
         return self
 
     def decimal(self, column, length=17, precision=6, nullable=False):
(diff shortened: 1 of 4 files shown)


josephmancuso:
Looking good so far! Can't wait to see how this turns out 👍

Marlysson:
@josephmancuso The work here seems to be done. There some choices that I did over the development that would be great if you validated them.

- Changes in create column method ( change to verify default value and to use **kwargs depending on which type of creating table query.
- The default value for temporal columns in MySQLGrammar.
- Was created the field **use_current_timestamp** in **Column** class (_I didn't find a reasonable place to put it._ To me, each grammar class should contain its default value for temporal fields ).
- Was created the default value for others column's type in **Column** class. ( Needed tests, but this PR isn't the scope. )

src/masonite/orm/blueprint/Blueprint.py
```diff
@@ -113,6 +122,19 @@ def datetime(self, column, nullable=False):
...
         return self
 
+    def timestamp(self, column, nullable=False):
+        self._last_column = self.new_column("timestamp", column, None, nullable)
+        self._columns += (self._last_column,)
+        return self
+
+    def timestamps(self):
+        created_at = self.new_column('timestamp', 'created_at', None, nullable=False).use_current()
+        updated_at = self.new_column('timestamp', 'updated_at', None, nullable=False).use_current()
```
josephmancuso, line 132: Should we add some of these options here in the parameters list? like 

```python
def timestamps(self, nullale=False, current=True):
    #..
```

This way we can also easily add the options as parameters. I think adding the ability to do that is simple enough

josephmancuso (changes requested):
Looks awesome. I'll merge this in a little later today.

Marlysson:
Ok. I will do.

I haver a doubt: Is it define the parameter nullable=[True|False] same that calling .nullable() method? Because we coul
ld be remove one argument from Column object.

josephmancuso:
I actually already did this and merged it

josephmancuso:
I also refactored a bit of the base grammar.

josephmancuso:
other than that, Great job! ❤️

josephmancuso:
> To update the updated_at field in MySQL there's the "ON UPDATE" to update the field by a trigger. Will this done by an update in python code or SQL statement?

it's a good question. I think with Python code for now like a global scope but we will see down the road

Marlysson:
You are on fire on this project haha.. I hope to contribute more here.

PR #45, opened 2020-03-02
Collection: is_empty & sum methods

ghost:


src/masonite/orm/collection/Collection.py
```diff
@@ -60,7 +60,7 @@ def implode(self):
         pass
 
     def is_empty(self):
-        pass
+        return True if not self.items else False
```
josephmancuso, line 63: You can pro just do `return not self.items` right?

josephmancuso (changes requested):

josephmancuso:
@sphinxxanxus I went ahead and just made the changes myself. They were minor. Great job!

PR #46, opened 2020-03-03
Schema: Add foreign key constraints

Marlysson:
**Indexes:** Create/Drop foreign indexes by name
- [ ] Add a new index named as linked tables[1]
- [ ] Drop indexes [2]

**Constraints:** Create/Drop indexes by name
- [ ] Create a foreign index indicating the linked tables[3]
- [ ] Drop foreign index [2]

**References:**

[1] https://orator-orm.com/docs/0.9/schema_builder.html#adding-indexes
[2] https://orator-orm.com/docs/0.9/schema_builder.html#dropping-indexes
[3] https://orator-orm.com/docs/0.9/schema_builder.html#foreign-keys

### DIFF
--- a/src/masonite/orm/blueprint/Blueprint.py
+++ b/src/masonite/orm/blueprint/Blueprint.py
@@ -1,3 +1,7 @@
+from ..relationships.ForeignKeyRelation import ForeignKeyRelation
+from ..schema.constraints.constraints import UniqueConstraint, ForeignKeyConstraint
+
+
 class Column:
     def __init__(
         self,
@@ -65,8 +69,7 @@ class Blueprint:
         self._last_column = None
         self._action = action
         self._default_string_length = default_string_length
-        self._last_relation = None
-        self. _relations = []
+        self._constraints = []
 
     def string(self, column, length=255, nullable=False):
         self._last_column = self.new_column("string", column, length, nullable)
@@ -196,32 +199,33 @@ class Blueprint:
     def unsigned_integer(self):
         pass
 
-    def foreign(self, local_column=None):
-        self._last_relation = {}
-        self._last_relation.update({"local_column": local_column})
-        return self
+    def foreign(self, *columns):
+        if not columns:
+            raise ValueError(
+                "You should define local key definitions on relationships."
+            )
 
-    def references(self, external_column=None):
-        self._last_relation.update({"external_column": external_column})
-        return self
+        columns_constraints = [
+            self.new_column(column_name, "unsigned", None) for column_name in columns
+        ]
 
-    def on(self, table=None):
-        self._last_relation.update({"table": table})
-        self._relations.append(self._last_relation)
-        self._last_relation = None
+        foreign_relation = ForeignKeyRelation(columns_constraints)
+        constraint = ForeignKeyConstraint(foreign_relation)
 
-        return self
+        self._constraints.append(constraint)
+
+        return foreign_relation
 
     def to_sql(self):
         if self._action == "create":
             return (
-                self.grammar(creates=self._columns, table=self.table)
+                self.grammar(creates=self._columns, table=self.table, constraints=self._constraints)
                 ._compile_create()
                 .to_sql()
             )
         else:
             return (
-                self.grammar(creates=self._columns, table=self.table)
+                self.grammar(creates=self._columns, table=self.table, constraints=self._constraints)
                 ._compile_alter()
                 .to_sql()
             )
@@ -242,6 +246,7 @@ class Blueprint:
 
     def unique(self):
         self._last_column.unique()
+        self._constraints.append(UniqueConstraint(self._last_column))
         return self
 
     def rename(self, old_column, new_column):
--- a/src/masonite/orm/blueprint/Blueprint.py
+++ b/src/masonite/orm/blueprint/Blueprint.py
@@ -18,11 +18,10 @@ class Column:
         self.constraint_type = None
         self.after_column = None
         self.old_column = ""
-        self.use_current_timestamp = False
         self.default = default
         self._action = action
         self._change = False
-
+        
     def nullable(self):
         self.is_null = True
         return self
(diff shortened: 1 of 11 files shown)


Marlysson:
@josephmancuso Does our *unsigned* method will have the different signature from others orms?

Instead: 

```
blueprint.integer('some_value').unsigned()
```

Will be, right ?

```
blueprint.unsigned('some_value')
blueprint.unsigned_integer('some_value')
```

josephmancuso:
@Marlysson I think both

josephmancuso:
These 2 ways:

```python
blueprint.integer('some_value').unsigned()
blueprint.unsigned_integer('some_value')
```

Marlysson:
@josephmancuso I've tried abstract the relations with some way but just came in my mind use of dictionary haha, do you have some other way?[1] For now, are working, now I will implement BaseGrammar to handle these relations and test the complete constraint.

[1] I started developing one class called **ForeignKeyRelation** but it just has 3 attributes and nothing more. I avoided this idea.

```python
class ForeignKeyRelation:

    # init method

    def foreign(self, local_column=None):
        self.local_column = local_column

    def references(self, external_column=None):
        self.external_column = external_column

    def on(self, table=None):
        self.foreign_table = table
```

[2] I think the use of **constraint_type = "foreign_key"** is too necessary ( to match with method constraint ), but I don't see where to put it on workflow.

tests/schema/mysql/test_mysql_foreign_key_constraint.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_correct_data_of_foreign_key_when_defined_more_than_one_relation(self):
+        pass
+
+    def test_throw_exception_when_foreign_key_column_doesnt_exist(self):
+        pass
+    
+    def test_throw_exception_when_columns_related_having_different_types(self):
+        pass
+
+    def test_throw_exception_when_foreign_key_table_doesnt_exist(self):
```
Marlysson, line 32: @josephmancuso Does this verification requires an integration test, right? Because it should verify if some table exists physically in the database...

josephmancuso: eventually probably but for now we can just assert the sql works. I would make a test somewhere to make sure the SQL is correct

tests/schema/mysql/test_mysql_schema.py
```diff
@@ -93,6 +93,14 @@ def test_column_exists(self):
...
         self.assertEqual(to_sql, sql)
 
+    def test_table_exists(self):
+        to_sql = self.schema.has_table("users", query_only=True)
+
+        sql = getattr(
+            self, inspect.currentframe().f_code.co_name.replace("test_", "")
+        )()
+        self.assertEqual(to_sql, sql)
+
```
josephmancuso, line 103: You'll need a test that looks like this:

```python
    def test_sets_foreign_constraint(self):
             table.string('name')
             table.integer('profile_id')

             table.foreign('profile_id').references('user').on('users')

        sql = getattr(
            self, inspect.currentframe().f_code.co_name.replace("test_", "")
        )()
        self.assertEqual(to_sql, sql)
```

and then make the child test like this:

```python
def sets_foreign_constraint(self):
     return (
        "CREATE TABLE `users` ("
            "`name` VARCHAR(255) NOT NULL, "
            "`profile_id` INT UNSIGNED,  "
                "INDEX profile_users_profile_id, "
                "FOREIGN KEY (`profile_id`)"
                "REFERENCES `profile`(`id`)"
        ")"
    )
```

src/masonite/orm/blueprint/Blueprint.py
```diff
@@ -196,16 +199,33 @@ def unsigned(self, column, length=None, nullable=False):
     def unsigned_integer(self):
         pass
 
+    def foreign(self, *columns):
```
josephmancuso, line 202: This definition looks wrong. It should only accept a column name:

```python
table.foreign('profile_id').references('profile').on('id')
```

josephmancuso (changes requested):

josephmancuso:
Is there a checkbox for you on the right side where it says something like "allow commits from repo owner"?

josephmancuso:
I'd like to work on this with you but by default idt I can push to this pull request

src/masonite/orm/blueprint/Blueprint.py
Marlysson: Ok. I did put *args because of this statement:

```sql
FOREIGN KEY (product_category, product_id)
      REFERENCES product(category, id)
```
Referencing two columns in the same constraint.

Marlysson:
It's checked, but with a different message. 
Sem título

tests/schema/mysql/test_mysql_schema.py
Marlysson: Ok, After that, I remove the exceptions that ensure fields are filled?
I will change the tests to this.

src/masonite/orm/blueprint/Blueprint.py
Marlysson: Here the reference: https://dev.mysql.com/doc/refman/8.0/en/create-table-foreign-keys.html#foreign-key-restrictions

The columns and table aren't quoted with `, I found it strange.

src/masonite/orm/grammar/mysql_grammar.py
```diff
@@ -146,12 +146,12 @@ def create_column_length(self):
     def unique_constraint_string(self):
         return "CONSTRAINT {clean_column}_unique UNIQUE ({clean_column})"
```
Marlysson, line 147: @corentinalcoy @josephmancuso 
Does here is just add a *args to receive more than one column?

ghost: No wait, Joe is looking the way the schema module is done. I think we have a problem...

josephmancuso:
Closing in favor of https://github.com/MasoniteFramework/orm/pull/77

PR #69, opened 2020-03-07
Collection: Finish implementation 

### LINKED ISSUE #31: Add Collection Methods

ghost:
- Finish Collection implementation
- Add test for each methods
fix #31

### DIFF
--- a/src/masonite/orm/collection/Collection.py
+++ b/src/masonite/orm/collection/Collection.py
@@ -255,7 +255,7 @@ class Collection:
         op = '=='
         value = args[0]
 
-        if len(args) == 2:
+        if len(args) >= 2:
             op = args[0]
             value = args[1]
 
--- a/src/masonite/orm/collection/Collection.py
+++ b/src/masonite/orm/collection/Collection.py
@@ -157,12 +157,19 @@ class Collection:
         self._items += items
         return self
 
-    def pluck(self, attribute):
-        attributes = []
+    def pluck(self, value, key=None):
+        if key:
+            attributes = {}
+
+        else:
+            attributes = []
         for item in self:
-            for key, value in item.items():
-                if key == attribute:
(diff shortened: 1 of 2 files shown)


ghost:
@josephmancuso I can't set you like a reviewer
Please review
Thanks

src/masonite/orm/collection/Collection.py
```diff
@@ -49,9 +54,14 @@ def collapse(self):
...
 
-    def contains(self):
-        # todo implements this methods
-        pass
+    def contains(self, key, value=None):
+        if value:
+            return self.contains(lambda x: self._data_get(x, key) == value)
+
+        if self._check_is_callable(key, raise_exception=False):
+            return self.first(key) is not None
```
josephmancuso, line 62: can this just e `return self.first()`?

src/masonite/orm/collection/Collection.py
```diff
@@ -251,16 +293,42 @@ def _get_value(self, key):
...
         if callable(value):
             return value()
         return value
 
-    def _check_is_callable(self, callback):
+    def _check_is_callable(self, callback, raise_exception=True):
         if not callable(callback):
+            if not raise_exception:
+                return False
             raise ValueError("The 'callback' should be a function")
```
josephmancuso, line 318: change this to should be a callable. It could be a classes method. function is a little wrong here even though a method is technically just a bound function

josephmancuso (changes requested):

src/masonite/orm/collection/Collection.py
ghost: In this case we need to return boolean.
I think self.first() is not the good response

josephmancuso: Ok

josephmancuso (approved):

PR #81, opened 2020-03-15
Feature/64 Implementing touch() method in Model.

### LINKED ISSUE #64: Add touching to model
This should simply set the updated at

Marlysson:
Close #64 
@josephmancuso I only update the "updated_at" attribute. 
But it's needed verify if model is created, if already exists some "updated_at" in "dirty" attribute?

To a preliminary implementation I've used datetime native type, should we use pendulum as date manipulation library?


Marlysson:
@josephmansuso

Should I use just ```updated_at = "now"``` instead set the datetime directly?

Besides tests, does lack something in implementation?

josephmancuso:
> Should I use just updated_at = "now" instead set the datetime directly?

I don't think that will actually work. I think i did make a test for it but I don't think its valid SQL.

This just needs a test and its good to go

Marlysson:
Is it the test: "save" a model and when "recover" it should returns with updated_at equals the set before?

josephmancuso:
not sure what you mean. Touch should just update the updated_at and save

josephmancuso:
pushed some changes and added a few tests

josephmancuso (approved):

PR #90, opened 2020-03-25
Refactor compile table statements

Marlysson:
- Drop **_compile_from** method in replacement to **_compile_table(self, table)**
    - Motivation: repeated lines

- Replacement some statements from **_compile_column** to **_compile_table** when was being formatted a table and vice-versa ( **compile_table** to **compile_column** when formatting a column )


josephmancuso (approved):

josephmancuso:
yes good idea

PR #91, opened 2020-03-26
Blueprint: Fix rename column function

### LINKED ISSUE #56: Add max, min, avg and sum tests

### LINKED ISSUE #85: Blueprint: Rename methods not working
Screenshot from 2020-03-18 01-19-02

@josephmancuso I think you should fix this. But If you want it I can.

corentinalcoy:
fix #85 
fix #56


corentinalcoy:
@josephmancuso

src/masonite/orm/grammar/BaseGrammar.py
```diff
@@ -109,7 +109,7 @@ def _compile_alter_columns(self):
             sql += getattr(self, "{}_column_string".format(column._action))().format(
                 column=self._compile_column(column.column_name),
                 old_column=self._compile_column(column.old_column),
-                data_type=self.type_map.get(column.column_type),
+                data_type=self.type_map.get(column.column_type) if self.type_map.get(column.column_type) else "",
```
josephmancuso: You don't need this ternary operator here. the `.get()` on a dictionary takes a second argument for a default. You can just do `self.type_map.get(column.column_type, "")`

josephmancuso (changes requested):

corentinalcoy:
Fix #56

josephmancuso:
fixed. The update command actually fires the query. So what you need to do for testing is pass in the `dry=True` parameter to not fire the query. Then you can assert it. See the update method now

tests/utils.py
```diff
@@ -0,0 +1,29 @@
+from unittest import mock
+
+from src.masonite.orm.connections.ConnectionFactory import ConnectionFactory
+from src.masonite.orm.connections.MySQLConnection import MySQLConnection
+from src.masonite.orm.connections.SQLiteConnection import SQLiteConnection
+
+
+class MockMySQLConnection(MySQLConnection):
```
josephmancuso, line 8: How does this mocking stuff work?

ghost: I do this to not fire the query.
It is a normal mocking of the `make_connection` function. 
This method will return a mock.Magick that help to not fire the query on the database really.

ghost:
> fixed. The update command actually fires the query. So what you need to do for testing is pass in the `dry=True` parameter to not fire the query. Then you can assert it. See the update method now

Let me check
I

ghost:
What do you update?
I don't see it

ghost:
@josephmancuso

tests/grammar/mysql/test_mysql_qmark.py
```diff
@@ -18,6 +18,15 @@ def test_can_compile_select(self):
...
+        mark = self.builder.where('name', 'Joe').delete()
+
+        sql, bindings = getattr(
+            self, inspect.currentframe().f_code.co_name.replace("test_", "")
+        )()
+        self.assertEqual(mark.to_qmark(), sql)
+        self.assertEqual(mark._bindings, bindings)
+
     def test_can_compile_update(self):
         mark = self.builder.update({"name": "Bob"}, dry=True).where("name", "Joe")
```
josephmancuso, line 31: @sphinxxanxus  here

ghost:
Okay.
It's good.
If it's looking good to you, please merge

josephmancuso:
Looks great!

PR #93, opened 2020-04-07
Add greater_than and less_than to the where method

### LINKED ISSUE #92: Add greater_than and less_than to the where method
We currently have something like this:

```python
builder.where('id', 10).get()
```

But we should ALSO be able to do any one of these:

```python
builder.where('id', '=', 10).get()
builder.where('id', '>, 10).get()
builder.where('id', '<', 10).get()
builder.where('id', '<=', 10).get()
builder.where('id', '>=, 10).get()
```

@Marlysson or @sphinxxanxus ?

corentinalcoy:
Fix #92

### DIFF
--- a/src/masonite/orm/builder/QueryBuilder.py
+++ b/src/masonite/orm/builder/QueryBuilder.py
@@ -1,7 +1,6 @@
 import inspect
 
 from ..collection.Collection import Collection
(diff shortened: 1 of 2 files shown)


corentinalcoy:
@josephmancuso 
I am done

josephmancuso (approved):
looks good to me!

PR #123, opened 2020-05-30
Adds hidden attribute to models

### LINKED ISSUE #112: Add __hidden__
Need to add a __hidden__ attribute to hide what gets serialized. This is done when hiding passwords or things that should not be seen.

josephmancuso:
Closes #112


josephmancuso:
This is used to hide things like password when serializing to a dictionary

PR #167, opened 2020-09-26
Feature/158 - Adds insert ID to the insert result set

josephmancuso:
By default, when you do an insert with most databases it won't actually return the result until you do a select. With the Python connection it usually returns an empty dictionary or an empty tuple. 

Because of this we have to just return whatever result we pass into the query builder (or model). 

So if we are creating a username and email:

```python
User.create({'username': "Joe", "email": "joe@joe.com"})
```

We have the results there really. So we're just going to return those:

```
results = {
    "username": "Joe",
    "email": "joe@joe.com"
}
```

A new issue comes up though. How do we get the primary key we just inserted?

SQLite is easy cuss its just an attribute on the connection cursor class.

PostgresSQL is also easy because we can append RETURNING * to the insert which will actually return the inserted result as it was inserted.

MySQL we will have to call a new SELECT LASTID() style query and use that result


josephmancuso:
@girardinsamuel @Marlysson Doesn't need an official review but worth a look

josephmancuso:
**STILL NEED TO ADD MYSQL SUPPORT FOR THIS**

Marlysson:
It's ok, I'll take a look.

src/masoniteorm/orm/query/processors/SQLitePostProcessor.py
```diff
@@ -0,0 +1,28 @@
+class SQLitePostProcessor:
+    """Post processor classes are responsable for modifying the result after a query.
+
+    Post Processors are called after the connection calls the database in the
+    Query Builder but before the result is returned in that builder method.
+
+    We can use this oppurtunity to get things like the inserted ID.
+
+    For the Postgres Post Processor we have a RETURNING * string in the insert so the result
```
girardinsamuel: Maybe it would make more sense to write only the connector specifics ?
- Write the Postgres Post Processor part in `PostgresPostProcessor`.
- Add SQLite, that we just need to get the`lastrowid` attribute here and same for MySQLPostProcessor

tests/sqlite/builder/test_sqlite_builder_insert.py
```diff
@@ -0,0 +1,42 @@
...
+            grammar=SQLiteGrammar,
+            connection=connection,
+            table=table,
+            # model=User,
+            connection_details=DATABASES,
+        ).on("sqlite")
+
+    def test_insert(self):
+        builder = self.get_builder()
+        print(
```
girardinsamuel: Here I guess you want to add an assert to check it the builder returned
````
{
     "id": 1,
     "name": "Joe",
      "email": "joe@masoniteproject.com",
      "password": "secret",
} 
```
?

src/masoniteorm/orm/query/QueryBuilder.py
```diff
@@ -315,15 +321,26 @@ def create(self, creates={}, query=False, **kwargs):
...
-        if self.owner:
-            return self.owner.hydrate(creates)
-        return creates
+        connection = self.new_connection()
+        result = connection.query(self.to_sql(), self._bindings, results=1)
+
+        if self._model:
+            id_key = self._model.get_primary_key()
+
+        processed_results = self.get_processor().process_insert_get_id(
```
girardinsamuel, line 336: it took me a bit of time to understand this part, maybe should we name `processed_results` / `results` consistently ?

With results=1 and not * I get that this will return one record, the one just created. So I am okay with `result` as returned value and then after doing post-processing should we keep `processed_result` without s ?

I find it more clear 👍

girardinsamuel (commented):
Here is a small review, you were right it's a great way to get into the project !

src/masoniteorm/orm/query/processors/SQLitePostProcessor.py
Marlysson: @girardinsamuel Just like a inheritance to create a base and change just what will be different between "type of processors" ?

girardinsamuel: I was just thinking about the docstrings 😛  but it could be applied to the real code too 👍 
Though for now it might not be necessary as there is not much there

josephmancuso: The docstring yes I need to update lol. I copied and pasted the classes and these followed

PR #168, opened 2020-09-27
added polymorphic relationship

### LINKED ISSUE #160: Add polymorphic relationships

josephmancuso:
Closes #160


josephmancuso:
This adds the ability to do a 1 to 1 polymorphic relationship. 

A polymorphic relationship is when a record on a table relates to 1 or more other types of table entities. So a user can like a post or like a comment or like a video. 

So a common polymorphic architecture is to have 1 likes table and then 3 post, comment and video tables.

The likes table relates to any one of those through the use of 2 columns: a type column and a record id column. Traditionally these are `record_type` and `record_id` by default but could commonly change to anything that makes more sense like `like_type` or `like_id`. This is entirely up to the developer or database engineer.

The way we are planning on doing this is we will require a "morph map". This is a relation between a record_type and a model. So on the table we save in the likes table `record_type = user` and `record_id = '1'`. This would link to the User model and fetch the first record.

The setting of the morph map is done through the 

```python
morph_to.set_morph_map({"user": User, "article": Articles})
```

class. This can be set before the relationships ever get called. If using Django it could be done inside the booting file or if Masonite it can be set into a service provider.

@girardinsamuel @Marlysson This is a good read to see how polymorphic relationship will work with Masonite and how easy it was to get this to work.

josephmancuso:
It could be possible to find a model without needing to set a morph map. Not sure how but that can be a quality of life feature down the road

josephmancuso:
I'm not gonna add the ability to eager load polumorphic relationships just yet. This is something that will be a little tough to figure out

PR #170, opened 2020-09-27
Feature/159 - Adds transactions

josephmancuso:
Adds transactions to:

- [x] Sqlite 
- [x] MySQL
- [x] Postgres

This PR only adds support for processing a transaction on an individual connection and not at a global level. A global level will require some more thoughts. Not sure if @girardinsamuel or @Marlysson have any thoughts on how we would do it at a global level. 

In Masonite tests we do something like:

```python
DB.begin_transaction()
```

and then everything after it will be inside a transaction and then either db.rollback or db.commit. but it needs to be set on the connection class so not sure where this global setting goes ..


josephmancuso:
@girardinsamuel @Marlysson worth a read

girardinsamuel:
Not sure what you mean by global level ? Could not we get the "current connection" used from the global `db` instance ? and then from the connection call the correct transaction methods ?

josephmancuso:
@girardinsamuel think of it in terms of a test case where we might have something like this:

```
# Start a transaction
# Do everything in here on multiple connections
# rollback the transaction
```

but I think there will be multiple connections opening and closing in the middle. I'm not sure how Orator handled that. or even if it did

josephmancuso:
one way I guess would be to set some global connection setting and then every connection we can open it already as a transaction but then not sure how to rollback all of them

PR #171, opened 2020-09-29
Clean/schema - Refactors and cleans Schema class

josephmancuso:
Changes how the schema class is constructed


josephmancuso:
Don't start any reviews yet. Still needs some work

PR #174, opened 2020-09-30
added connection resolver class

josephmancuso:
Adds a connection resolver class. This class will be used to "resolve" (or find) the connection settings. This class will need to be called sometime before connections start.

for Masonite we can likely just throw this in a service provider.

For your own projects outside of Masonite we can throw this inside a settings class or wherever you want to store the connections dictionary.

### DIFF
--- a/cc.py
+++ b/cc.py
@@ -1,11 +1,10 @@
 from src.masoniteorm.query import QueryBuilder
 from src.masoniteorm.connections import MySQLConnection, PostgresConnection
 from src.masoniteorm.query.grammars import MySQLGrammar, PostgresGrammar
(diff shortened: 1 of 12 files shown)


josephmancuso:
@girardinsamuel here is that connection resolver concept I was talking about

girardinsamuel:
Nice 👍  ! 
[UPDATED] forgot about old comment.

girardinsamuel:
If I understand correctly the DatabaseServiceProvider in Masonite will call the `ConnectionResolver` to `register` all connections defined in `config/database.py` ?

josephmancuso:
that's correct

Issue #172, opened 2020-09-29
migration fails on sqlite


josephmancuso:
```python
        with self.schema.table('videos') as table:
            table.unsigned_integer('playlist_id').nullable()
            table.foreign('playlist_id').references('id').on('playlists')
```

```
ALTER TABLE "videos" ADD "playlist_id" UNSIGNED INT, ADD CONSTRAINT videos_playlist_id_foreign FOREIGN KEY ("playlist_id") REFERENCES "playlists"("id") 
```

```
near ",": syntax error
```

Marlysson:
@josephmancuso Can I take this?
Investigating about this, I found that we should enable foreign key support in sqlite databases through  "PRAGMA foreign_keys = ON". This would be done in each opened connection.

https://sqlite.org/foreignkeys.html

josephmancuso:
oh you know what, you are actually right. I thought it was more complex than that.

We might be able to just straight up put it in the alter table method in the grammar class

Marlysson:
@josephmancuso The error about ```near ",": syntax error``` is that sqlite doesn't support add constraints in a table through ```ALTER TABLE```. The recommended is defined it when create the table ( with CREATE TABLE ) , but the case of alter table to add new foreign keys is so common that this suggestion don't should be considered.

What should we do?

Here is the reference: https://www.sqlite.org/omitted.html

josephmancuso:
@Marlysson in the SQLite grammar class we just need to add that 

```python
def alter_format(self):
        return "PRAGMA foreign_keys = ON; ALTER TABLE {table} {columns}{constraints}{foreign_keys}"
```

josephmancuso:
we might need to also put it back off at the end of the query?

Marlysson:
Yes, the problem is when the query uses ```alter table``` together ```add constraint``` the sqlite doesn't support.
But I'll try what you suggested.

By that I read, it's just run once on connection and the sqlite database is supported with foreign keys.

josephmancuso:
oh hmm ok

Marlysson:
Hmm, by changing in the grammars, all tests should be changed too ( 19 more exactly until now ).

Is recommended to set that once in connection ( when connect to database maybe) ?

I think this bug is more "complex" than I've found because:

> The ALTER command in SQLite3 is limited to adding columns or renaming tables, so you cannot do this. Instead, you must re-create the table with the desired constraints.

[...]

PR #176, opened 2020-09-30
Fix/172 - Alter table to add constraints in a existent table - sqlite.

Marlysson:
Once that sqlite doesn't support define just a constraint in alter table statement there were some tests that set up just the foreign key constraint.

So, all times of alter table to add constraint ( of foreign key type ) should be set the new column that it are related to. I think it's. 

@josephmancuso please can you review this and if make sense?


Marlysson:
**IN DRAFT**

Marlysson:
@josephmancuso Done. 

Maybe need verify if wasn't setted foreign_keys to ON . But would be 2 queries just to do this. Then always I define it to True.
Tests: There's currently some way to do a integration tests?

Closes #172

src/masoniteorm/connections/SQLiteConnection.py
```diff
@@ -34,10 +34,12 @@ def __init__(
 
     def make_connection(self):
         """This sets the connection on the connection class"""
-        print("db", self.database)
 
         self._connection = sqlite3.connect(self.database, isolation_level=None)
 
+        self._cursor = self._connection.cursor()
+        self._cursor.execute("PRAGMA foreign_keys = ON", ())
```
josephmancuso, line 41: This should only be done when modifying the schema. We shouldn't turn pragma keys on every time we make a connection here

src/masoniteorm/connections/SQLiteConnection.py
```diff
@@ -111,7 +113,7 @@ def query(self, query, bindings, results="*"):
             dict|None -- Returns a dictionary of results or None
         """
         query = query.replace("'?'", "?")
-        print("running query: ", self, query, bindings)
```
josephmancuso, line 114: this is odd but this should stay for now. It helps me see what queries are running while testing

tests/sqlite/schema/test_sqlite_alter_table.py
```diff
@@ -61,29 +65,53 @@ def test_can_alter_index(self):
...
+        )()
+        self.assertEqual(blueprint.to_sql(), sql)
+
     def test_can_alter_foreign_key(self):
         with self.schema.table("users") as blueprint:
+            blueprint.unsigned_integer("profile_id").nullable()
             blueprint.foreign("profile_id").references("id").on("profile")
 
-        sql = """ALTER TABLE "users" ADD CONSTRAINT users_profile_id_foreign FOREIGN KEY ("profile_id") REFERENCES "profile"("id")"""
+        sql = """ALTER TABLE "users" ADD "profile_id" INTEGER, CONSTRAINT users_profile_id_foreign REFERENCES "profile"("id")"""
```
josephmancuso, line 94: I think these should have the pragma key string in them

josephmancuso (changes requested):
I think the best way to go about this is to override the _compile_alter_table on the sqlite grammar class and do something like this:

```python
    def _compile_alter(self):
        """Compiles a query for altering table schemas.

        Returns:
            self
        """
        self._sql = "PRAGMA foreign_keys = ON; "
        self._sql += self.alter_format().format(
            table=self._compile_table(self.table),
            columns=self._compile_alter_columns(),
            constraints=self._compile_alter_constraints(),
            foreign_keys=self._compile_alter_foreign_keys(),
        )

        self._sql = self._sql.rstrip(", ")

        self._sql += "; PRAGMA foreign_keys = OFF; "



        return self
```

@girardinsamuel @Marlysson This way it only happens for sqlite grammars, only for alters and if we really need to only do this when adding constraints we can do an `if` check if the grammar has any constraints, then prepend and append the pragma thing

josephmancuso:
I don't really understand this pragma thing actually. I'm reading it and it says each connection should turn pragma keys to on before executing a query in order to ensure foreign keys are honored? that sounds really dumb lol.

josephmancuso:
I also don't think this actually works. When I run it in my client it just says query: OK but nothing happens

josephmancuso:
I think the only way to do this is to actually run a bunch of queries. Like this:

```
https://stackoverflow.com/a/28869056
```

src/masoniteorm/connections/SQLiteConnection.py
Marlysson: You're right. I'll move to sqlite grammar.

josephmancuso:
I think what we will have to do is fetch all the columns on the table, build a new grammar class out of them except this time add the new foreign key references and then drop the old table ....

Marlysson:
> I don't really understand this pragma thing actually. I'm reading it and it says each connection should turn pragma keys to on before executing a query in order to ensure foreign keys are honored? that sounds really dumb lol.

Me neither. But seems that was problems with backwards compatibilities. Started with some behavior and its uses grow up so fast to possible changes.

> I also don't think this actually works. When I run it in my client it just says query: OK but nothing happens

I really need to do integration tests before push the changes..
But what operation do you tried?

Marlysson:
> I think what we will have to do is fetch all the columns on the table, build a new grammar class out of them except this time add the new foreign key references and then drop the old table ....

Build a backup before add foreign keys or just copy table infos?

Marlysson:
@josephmancuso based on what you said in twitter, this feature will by putted in stand by until conclude the stuffs in your tweet?

josephmancuso:
@Marlysson I'm not sure. I think so? idt what's implemented in here actually works. At least it doesn't work for me. Whenever I try to run this:

```
ALTER TABLE "users" ADD "profile_id" INTEGER, CONSTRAINT users_profile_id_foreign REFERENCES "profile"("id")
```

It doesn't actually seem to work for me

josephmancuso:
when I run it in table plus it doesn't seem to actually add the constraint

Marlysson:
I've tried here in table plus too and it just create the relationship, but not the constraint.
You said that will take a look in changes in DLL SQL , are we try the snippet that you suggest in thread or wait to new solutions?

josephmancuso:
@Marlysson take a look at some of these tests https://github.com/doctrine/dbal/blob/16878650d79934c31e27a13618ed6f3effee80d8/tests/Doctrine/Tests/DBAL/Platforms/SqlitePlatformTest.php#L297

Marlysson:
@josephmancuso I'll put this PR in stand by because you are trying to redefine schema/grammar logic.

But I'm paying attention.

josephmancuso (approved):

josephmancuso:
I love it but I will close in favor of #181

PR #177, opened 2020-10-01
Add ORMProvider

girardinsamuel:
* Add ORMProvider for use of commands in Masonite
* Fix connections when running migrations by removing old code

### DIFF
--- a/src/masoniteorm/connections/MySQLConnection.py
+++ b/src/masoniteorm/connections/MySQLConnection.py
@@ -69,9 +69,8 @@ class MySQLConnection(BaseConnection):
     def get_default_query_grammar(cls):
         return MySQLGrammar
 
-    @classmethod
     def get_database_name(self):
-        return self().database
+        return self.database
 
     def commit(self):
         """Transaction"""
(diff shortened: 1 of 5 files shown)


src/masoniteorm/connections/MySQLConnection.py
```diff
@@ -69,28 +69,9 @@ def reconnect(self):
...
-        connection_details.setdefault("port", int(self.connection_details.get("port")))
-        connection_details.setdefault("db", self.connection_details.get("database"))
-        connection_details.update(self.connection_details.get("options", {}))
-
-        return connection_details
-
     @classmethod
     def get_database_name(self):
-        return self().get_connection_details().get("db")
+        return self().database
```
josephmancuso: I don't believe this will work the way you think it will work. This is a class method. class methods pass in the class into the first variable instead of the instance. So in this example this will always return false because you are instantiating a new class with no parameters.

It should just be 

```python
def get_database_name(self):
    return self.database
```

josephmancuso (changes requested):

src/masoniteorm/connections/MySQLConnection.py
girardinsamuel: Yes, that's what I had done first. Then as I wasn't sure I switch back (both are working maybe because self() returned the already instantiated class ? like a singleton ?).

Anyway I made requested changes !

girardinsamuel:
I added `seed:run` command.
Running those commands works for me in the demo project 👍  ! Finally !

josephmancuso (approved):

PR #181, opened 2020-10-07
Rewrite/schema rewrite

josephmancuso:
@Marlysson @girardinsamuel this is just the very very very first start of what I'm working on but I'm going to work through this solution and see where it leads me. This is how it looks like other ORMS do it like Doctrine (PHP) and alchemy does it


josephmancuso:
these tests are all gonna fail until I work through the PR

josephmancuso:
Most notable differences are when you look at the new platform class, as well as the new table and table diff classes

src/masoniteorm/schema/platforms/SQLitePlatform.py
```diff
@@ -0,0 +1,143 @@
...
+    def compile_create_sql(self, table):
+        columns = self.columnize(table.get_added_columns())
+        
+        return self.create_format().format(
+            table=self.table_string().format(table=table.name).strip(),
+            columns=', '.join(self.columnize(table.get_added_columns())).strip(),
+            constraints=', ' + ', '.join(self.constraintize(table.get_added_constraints())) if table.get_added_constraints() else ""
+        )
+
+    def compile_alter_sql(self, diff):
```
josephmancuso, line 56: @Marlysson @girardinsamuel I think writing the methods like this has been much easier with the table and table diff structure

src/masoniteorm/schema/platforms/SQLitePlatform.py
```diff
@@ -0,0 +1,143 @@
...
+        
+        if diff.renamed_columns:
+            sql.append("CREATE TEMPORARY TABLE __temp__{table} AS SELECT {original_column_names} FROM {table}".format(
+                table=diff.name,
+                original_column_names=', '.join(diff.from_table.added_columns.keys())
+            ))
+            sql.append("DROP TABLE {table}".format(table=diff.name))
+
+            columns = diff.from_table.added_columns
+            columns.update(diff.renamed_columns)
```
josephmancuso, line 91: with the table and table diff structure we can do something cool like this where we have the columns added on the table and then we can overwrite them with the renamed columns. Nice little short piece of effective code.

tests/sqlite/schema/test_table_diff.py
```diff
@@ -0,0 +1,119 @@
...
+        diff = TableDiff("users")
+        diff.from_table = table
+        diff.rename_column("post", "comment", "integer")
+
+        sql = [
+            "CREATE TEMPORARY TABLE __temp__users AS SELECT post, user, email FROM users", 
+            "DROP TABLE users",
+            'CREATE TABLE "users" (comment INTEGER, user INTEGER, email INTEGER)',
+            'INSERT INTO "users" (comment, user, email) SELECT post, user, email FROM __temp__users',
+            'DROP TABLE __temp__users',
```
josephmancuso: @Marlysson @girardinsamuel with the table and table diff structure its really simple to create some complex queries required for sqlite to work

josephmancuso (commented):
@Marlysson @girardinsamuel added some comments here worth a look at. Not ready yet for a review but might be worth seeing how we're gonna do this now

src/masoniteorm/schema/Table.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def get_added_columns(self):
+        return self.added_columns
+
+    def rename_column(self, column, to):
+        pass
+
+    def set_primary_key(self, key):
+        self.primary_key = key
+        self.added_columns[key].set_as_primary()
```
Marlysson: The value "key" should be named as "column_name", right? Indicating that is a column name and is possible access it by its name. Was a bit difficult identify that it's a column.

tests/sqlite/schema/test_table_diff.py
```diff
@@ -0,0 +1,24 @@
...
+        table.add_column("name", "string")
+        table.add_constraint("name", "unique")
+
+        diff = TableDiff("users")
+        diff.from_table = table
+        diff.new_name = "clients"
+        diff.remove_constraint("name")
+
+        sql = ["ALTER TABLE users RENAME TO clients"]
+
```
Marlysson, line 21: Should constraint changes appears in sql?

Marlysson:
@josephmancuso That new structure until now it's much easier to understand.

josephmancuso:
@Marlysson perfect. That's exactly what we're going for

src/masoniteorm/schema/Table.py
josephmancuso: yup, totally agree

tests/sqlite/schema/test_table_diff.py
josephmancuso: no not in this case because the only diff is the renaming of the table

josephmancuso:
@Marlysson @girardinsamuel worked through this more. Got the tests to start passing by removing the old code. Will slowly work through adding them all back in this new structure. If you look at some of the latest commits you can see some cool things we can do now.

The new structure really cleans up a lot of the dirty code.

src/masoniteorm/schema/grammars/SQLiteGrammar.py
```diff
@@ -70,172 +70,6 @@ class SQLiteGrammar(BaseGrammar):
 
     null_map = {False: "NOT NULL", True: "NULL"}
 
-    """
```
Marlysson: Do all this code was removed in replacement by new platform structure? :o

josephmancuso: yes this is in the middle of a redesign still but the grammar classes will now no longer be responsible for compiling queries. This is now the job of the platform class

josephmancuso: I removed all the tests for alter tables for all drivers and I will be adding it all back using the new structure.

PR #184, opened 2020-10-10
MySQL, Postgres and SQLite Platform Schema

josephmancuso:
@Marlysson is this something you want to try out? Just need to start building out tests for the mysql platform class. It will be basically identical to the sqlite platform tests and structure


tests/mysql/schema/test_mysql_schema_builder.py
```diff
@@ -0,0 +1,105 @@
+import unittest
```
josephmancuso, line 1: @Marlysson this file is just building out the schema structures and then asserting the schema SQL

josephmancuso: Can just copy and paste from schema file and then convert to mysql structure

src/masoniteorm/schema/platforms/Platform.py
```diff
@@ -0,0 +1,56 @@
+class Platform:
```
josephmancuso, line 1: @Marlysson anything that is the same across platforms can go here

src/masoniteorm/schema/platforms/MySQLPlatform.py
```diff
@@ -0,0 +1,82 @@
+from .Platform import Platform
```
josephmancuso, line 1: @Marlysson This is the new platform classes responsible for compiling the SQL

josephmancuso (commented):
@Marlysson here are some comments if you want to work on the PR. Can do it right on the feature/182 branch. 

If you want to work on this it will free up my time to start the Postgres implementation

josephmancuso:
Building a new database requires:

* Platform class
* Connection class

Marlysson:
@josephmancuso 
Yep, I can do that. 

Anything difficult I can see your postgres parralel implementation.

josephmancuso:
@Marlysson yup I already started that. The 2 implementations will be very similar. MySQL will be easier because there are (I think) less tricks with MySQL. MySQL seems to work the way you would expect. I think. 

I started Postgres already so I'll open a PR on that soon

josephmancuso:
@Marlysson actually nvm I forgot I just did it in the branch because I made some parent classes I didn't want to duplicate. So you can just check Postgres in this branch

src/masoniteorm/schema/grammars/PostgresGrammar.py
```diff
@@ -94,53 +94,6 @@ class PostgresGrammar(BaseGrammar):
         "not_null": " NOT NULL ",
     }
 
-    def select_format(self):
```
Marlysson, line 97: These removed lines will just moved to platform or removed definitelly?

josephmancuso:
@Marlysson there are 2 almost the same grammar classes. For this PR if you need to pull anything from grammar pull it from the one in the `schema` directory and not the `query` directory. The `query` directory is the one being used for normal DML style queries (select, update, delete, etc)

josephmancuso:
@Marlysson I don't actually see a need for the schema grammars so they might be deleted after

josephmancuso:
actually added Python 3.9 to the tests

josephmancuso:
@Marlysson after coding all day the sqlite and Postgres implementations are done 👍 . Let me know if you need help on the MySQL implementation

josephmancuso:
@Marlysson sorry I had to test stuff so I worked on most of the MySQL platform already

Marlysson:
@josephmancuso 
 Sorry for late implementation. I'm still analysing to implement something. But keep going. I'm here seeing the commits and code for future needed.

josephmancuso:
@Marlysson at this point I THINK its done. I'm still reviewing it all. My migrations from masonite casts which is like 7 tables runs great on all 3 databases

josephmancuso:
I'm gonna merge this and start writing docs and getting another  beta version out

PR #192, opened 2020-10-16
added improved query statements

### LINKED ISSUE #16: experiment with emitting some kind of listeners
We need to be able to emit (maybe using python logger) and have listeners so we can build integrations that do things like show the query and the time they took to execute

josephmancuso:
Closes #16


josephmancuso:
Allows us to listener on the logger class for queries executed. We are able to get the query, the bindings and the query time now

PR #193, opened 2020-10-16
Adding MSSQL Support

### LINKED ISSUE #22: Add MSSQL Support
Need to wait for MySQL grammar to finish

josephmancuso:
Closes #22


josephmancuso:
@Marlysson @girardinsamuel this only took me like 3 days to do based on the current data structures but I think its pretty much done. What I have to do now is just double check the tests I'm asserting against a real Microsoft database

src/masoniteorm/connections/MSSQLConnection.py
```diff
@@ -0,0 +1,146 @@
...
+from .BaseConnection import BaseConnection
+from ..query.grammars import MSSQLGrammar
+from ..schema.platforms import MSSQLPlatform
+
+
+CONNECTION_POOL = []
+
+
+class MSSQLConnection(BaseConnection):
+    """Postgres Connection class."""
```
girardinsamuel: Don't forget to update the docstring here
```suggestion
    """MSSQL Connection class."""
```

src/masoniteorm/connections/MSSQLConnection.py
```diff
@@ -0,0 +1,146 @@
...
+        """This sets the connection on the connection class"""
+        try:
+            import pyodbc
+        except ModuleNotFoundError:
+            raise DriverNotFound(
+                "You must have the 'pyodbc' package installed to make a connection to Microsoft SQL Server. Please install it using 'pip install pyodbc'"
+            )
+
+        self._connection = pyodbc.connect(
+            f"DRIVER={'ODBC Driver 17 for SQL Server'};SERVER={self.host},{self.port};DATABASE={self.database};UID={self.user};PWD={self.password}",
```
girardinsamuel: Shouldn't this hardcoded DRIVER string be defined from config ? I don't know MSSQL very well so might be wrong.. . I guess it's also linked to the version of `pyodbc` we are using ?

src/masoniteorm/connections/MSSQLConnection.py
```diff
@@ -0,0 +1,146 @@
...
+    def commit(self):
+        """Transaction"""
+        if self.get_transaction_level() == 1:
+            self._connection.commit()
+            self._connection.autocommit = True
+
+        self.transaction_level -= 1
+
+    def begin(self):
+        """Postgres Transaction"""
```
girardinsamuel:
```suggestion
        """MSSQL Transaction"""
```

src/masoniteorm/schema/platforms/MSSQLPlatform.py
```diff
@@ -0,0 +1,233 @@
...
+        return sql
+
+    def add_column_string(self):
+        return "{name} {data_type}{length}"
+
+    def drop_column_string(self):
+        return "{name}"
+
+    def rename_column_string(self, table, old, new):
+        return f"EXEC sp_rename 'users.post', 'comment', 'COLUMN';"
```
girardinsamuel: Guess this first testing line can be removed
```suggestion
```

girardinsamuel (commented):
@josephmancuso Made a quick review 👍

src/masoniteorm/schema/platforms/MSSQLPlatform.py
josephmancuso: The first line is correct. the second line needs to be removed :)

src/masoniteorm/connections/MSSQLConnection.py
josephmancuso: You are right. We're gonna need a way to pass all the other options into the connection class. Right now we just pass the host, username, password, etc

josephmancuso:
This at least has feature parity with the other databases so will merge. Will also look into git store to store this mssql driver

PR #196, opened 2020-10-18
Feature/195

### LINKED ISSUE #195: Make a connector class
There really should be a single integration point with all these connections and databases. Right now we need to register things in the ConnectionFactory, the GrammarFactory and the Platform Factory.

There really should be something like this:

```python
ConnectionResolver.add(SQLiteConnector())
```

josephmancuso:
This adds a central point for connections. The connection class has the grammar and platform class so the connection class is really the central brain of the database integration. If we register the connection with the connection factory it will easily complete the connection point

Closes #195


josephmancuso:
This PR also removes all of the factories besides the connection factory

Issue #122, opened 2020-05-27
Be compatible to use UUID as primary key


rfschubert:
As per default in many databases, we should be able to use `uuid` as `pk`.

On `postgres` databases it is common to use `uuid` and not `int` as default `pk`. But is not forced users to use `uuid` so it may be easy to set which type of field we want to use.

`Django` currently forces user to declare the desired `id` and overrides the default, but need be done in every table, and it is quite annoying.

```python
from django.db import models

class MyTable(models.Model):
    id = models.UUIDField(default=uuid.uuid4, primary_key=True, editable=False)


class MyAnotherTable(models.Model):
    id = models.UUIDField(default=uuid.uuid4, primary_key=True, editable=False)


class MyAnotherTableAgain(models.Model):
    id = models.UUIDField(default=uuid.uuid4, primary_key=True, editable=False)
```

An idea is to use global config variable, as one way to override, but user still may be able to declare by itself.

*Sample:*
```python
OVERRIDE_DEFAULT_ID_FIELD_TYPE = "uuid"
```

I know use cases where system have almost all tables using `uuid` field as default and only 1 table with `int` as `pk` because of specific law pattern of one data that must be stored, so others users will also want to have `auto increment` on few tables.

josephmancuso:
oh i like this

josephmancuso:
so you're saying when you do 

```
table.increments('something')
```

it actually can map to a UUID?

rfschubert:
I don't see currently do `table.increments('something')` as useful, but for sample:

I'm using an table to generate sequential numbers because I need to use it on a gov database document.

So I have only one table on my mainframe like this:

```python
class MyIntegerTable(models.Model):
    pass
```
Because for default `Django` will use `id` as auto increment.

And also I have others `85~90` tables using `uuid` as primary key... so I have dozens of lines repeatedly on my projects with
```
id = models.UUIDField(default=uuid.uuid4, primary_key=True, editable=False)
```

But I think it was more simple to tell the ORM that I want to use for default `uuid` and only use `int` as id if I need. So I would have almost all servers without the need to manually specify that `pk` should be `uuid` in every table, and only one line telling to my ORM that my `MyIntegerTable` should have `int` as `pk`.

rfschubert:
I think that if I create a new table, It can generate the `id` field by itself, as every table should have an `id` field as `pk`.

[...]

PR #197, opened 2020-10-18
Be able to define global primary key type

girardinsamuel:
Quick try to implements #122 

@josephmancuso @hamzzy
Is something like this making sense ? doable ?

### DIFF
--- a/src/masoniteorm/query/QueryBuilder.py
+++ b/src/masoniteorm/query/QueryBuilder.py
@@ -1,5 +1,5 @@
 import inspect
-
+import uuid
 from ..collection.Collection import Collection
 from ..expressions.expressions import (
     SubGroupExpression,
@@ -349,6 +349,13 @@ class QueryBuilder:
             creates = kwargs
 
         self.set_action("insert")
+
+        # before insertion generates new uuid if ORM_PK_TYPE is UUID
+        ORM_PK_TYPE = "uuid"  # to be defined in configuration
+        if ORM_PK_TYPE == "uuid":
+            new_pk = uuid.uuid4()
+            self._creates.update({id_key: new_pk})
+
         self._creates.update(creates)
         if query:
             return self
--- a/src/masoniteorm/query/QueryBuilder.py
+++ b/src/masoniteorm/query/QueryBuilder.py
@@ -1,5 +1,4 @@
 import inspect
-import uuid
 from ..collection.Collection import Collection
 from ..expressions.expressions import (
     SubGroupExpression,
@@ -12,7 +11,7 @@ from ..expressions.expressions import (
     HavingExpression,
 )
 
-from ..scopes import BaseScope
+from ..scopes import BaseScope, UUIDPrimaryKeyScope
 
 from ..schema import Schema
 
@@ -349,6 +348,11 @@ class QueryBuilder:
             creates = kwargs
 
         self.set_action("insert")
+
+        # here is an not a right place to do that but somewhere we should do this:
+        # TODO: if pk column is of type uuid, (how?)
+        # then add scope
+        self.set_global_scope(UUIDPrimaryKeyScope())
 
         self._creates.update(creates)
         if query:
--- a/src/masoniteorm/query/QueryBuilder.py
+++ b/src/masoniteorm/query/QueryBuilder.py
@@ -350,12 +350,6 @@ class QueryBuilder:
 
         self.set_action("insert")
 
-        # before insertion generates new uuid if ORM_PK_TYPE is UUID
-        ORM_PK_TYPE = "uuid"  # to be defined in configuration
-        if ORM_PK_TYPE == "uuid":
-            new_pk = uuid.uuid4()
-            self._creates.update({id_key: new_pk})
-
         self._creates.update(creates)
         if query:
             return self
--- a/src/masoniteorm/query/QueryBuilder.py
+++ b/src/masoniteorm/query/QueryBuilder.py
@@ -11,7 +11,7 @@ from ..expressions.expressions import (
     HavingExpression,
 )
 
-from ..scopes import BaseScope, UUIDPrimaryKeyScope
+from ..scopes import BaseScope
 
 from ..schema import Schema
 
@@ -348,7 +348,6 @@ class QueryBuilder:
             creates = kwargs
 
         self.set_action("insert")
-
         self._creates.update(creates)
         if query:
             return self
--- a/src/masoniteorm/query/QueryBuilder.py
+++ b/src/masoniteorm/query/QueryBuilder.py
@@ -1,4 +1,5 @@
 import inspect
+
 from ..collection.Collection import Collection
 from ..expressions.expressions import (
     SubGroupExpression,
--- a/src/masoniteorm/query/QueryBuilder.py
+++ b/src/masoniteorm/query/QueryBuilder.py
@@ -349,11 +349,6 @@ class QueryBuilder:
 
         self.set_action("insert")
 
-        # here is an not a right place to do that but somewhere we should do this:
-        # TODO: if pk column is of type uuid, (how?)
-        # then add scope
-        self.set_global_scope(UUIDPrimaryKeyScope())
-
         self._creates.update(creates)
         if query:
             return self
--- a/src/masoniteorm/schema/Blueprint.py
+++ b/src/masoniteorm/schema/Blueprint.py
@@ -85,7 +85,7 @@ class Blueprint:
         )
         return self
 
-    def uuid(self, column, nullable=False):
+    def uuid(self, column, nullable=False, length=36):
         """Sets a column to be the UUID4 representation for the table.
 
         Arguments:
@@ -97,7 +97,7 @@ class Blueprint:
         Returns:
             self
         """
-        self._last_column = self.table.add_column(column, "uuid", nullable=nullable)
+        self._last_column = self.table.add_column(column, "uuid", nullable=nullable, length=length)
         return self
 
     def big_increments(self, column, nullable=False):
--- a/src/masoniteorm/schema/Blueprint.py
+++ b/src/masoniteorm/schema/Blueprint.py
@@ -85,8 +85,8 @@ class Blueprint:
         )
         return self
 
-    def uuid_increments(self, column, nullable=False):
-        """Sets a column to be the UUID primary key representation for the table.
+    def uuid(self, column, nullable=False):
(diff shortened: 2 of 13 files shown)


josephmancuso:
I'm not entirely sure what that ticket is asking for to be honest. Either way I don't think this would be the correct approach here. I think the way to go would be do make a global scope and inherit it on the model.

josephmancuso:
I think the scope would look something like this:

```python
class UUIDPrimary(BaseScope):
    """Global scope class to add uuid primary key"""

    def on_boot(self, builder):
        builder.set_global_scope(
            "_uuid_primary", self.set_uuid_create, action="insert"
        )

    def on_remove(self, builder):
        pass

    def set_uuid_create(self, builder):
        builder._creates.update(
            {
                builder._model.__primary_key__: uuid4(), # On insert, generate a uuid4 key right before it will try to create this record
            }
        )
```

then make a mixin to register it to the model and then you can have something like this:

@rfschubert would a syntax like this make this easier for you?

```python
class Client(Model, SoftDeletesMixin, UUIDPrimaryMixin, ...):
    pass
```

girardinsamuel:
Yeah it looks cleaner with a mixin ! But we have to add the mixin to every model though. Definitely doable because it is not very complex !

josephmancuso:
@girardinsamuel yes or have a base model and just have all your models inherit that model. Maybe that's not such a terrible idea to have that right out of the box with Masonite.

I'm just not sure if this achieves @rfschubert goal.

girardinsamuel:
We will wait for his answer but as a Django user I've come across the same problem of rewriting this uuid line and for me this solves the problem

josephmancuso:
@girardinsamuel oh perfect. Yeah I think it would work too but I thought maybe I didn't understand the original feature request. Ok if you think the global scope approach is good I don't see why we shouldn't make it :) I think it would be a cool scope even if it doesn't solve the problem lol. we might need to support more than just the primary key as guid though. Maybe we do something like `__uuids__ = ['object_id', 'transaction_id']` and we just loop through these and create uuids for each. Default to uuid4 but can also set an option in `__uuid_version__ = '4'`. Once everything is set we can put all the if checks in the `set_uuid_create` method of the scope

girardinsamuel:
@josephmancuso Like this idea, would be more flexible 👍 !
But from a Django perspective, it's a bit strange for me not be able to do this in the migrations :

```python
            table.uuid('id').primary()  # uuid as PK
            table.string('name')
            table.uuid('transaction_id') # other uuid field
            # or table.uuid3('transaction_id') ...
```

Instead we need to do if I got it correctly:

```python
# User.py
class User(Model, UUIDFieldsMixin):
    __uuids__ = ['id', 'transaction_id']
    __uuid_version__ = 4

# create_user_table_migration.py
            table.char('id')  # ?
            table.string('name')
            table.char('transaction_id') # ?
```
Okay for me too but as I am used to "the first one in Django", it's not intuitive.

josephmancuso:
@girardinsamuel yeah I agree. We can add support for a uuid in the schema like you have in this PR. That's fine for me. We would have to map uuid type to CHAR and then set the length default to 36 in the uuid schema method so that something like CHAR(36) will be used when you do table.uuid(..)

girardinsamuel:
Okay cool, I will try to complete and clean up this PR (and add some tests !) Time to put hands on the ORM code 😄

src/masoniteorm/schema/platforms/MySQLPlatform.py
```diff
@@ -15,7 +15,7 @@ class MySQLPlatform(Platform):
         "small_integer": "SMALLINT",
         "medium_integer": "MEDIUMINT",
         "increments": "INT UNSIGNED AUTO_INCREMENT PRIMARY KEY",
-        "uuid_increments": "CHAR PRIMARY KEY",
+        "uuid": "CHAR(36)",
```
josephmancuso: This should be CHAR. I would not hard code a length here

src/masoniteorm/schema/Blueprint.py
```diff
@@ -428,7 +443,7 @@ def primary(self, column):
         Returns:
             self
         """
-        self.table.set_primary_key([column])
+        self.table.set_primary_key(column)
```
girardinsamuel, line 446: @josephmancuso This was causing a bug so I fixed it, is it correct ?

src/masoniteorm/schema/platforms/PostgresPlatform.py
```diff
@@ -15,6 +15,7 @@ class PostgresPlatform(Platform):
         "small_integer": "SMALLINT",
         "medium_integer": "MEDIUMINT",
         "increments": "SERIAL UNIQUE",
+        "uuid": "CHAR(36)",
```
josephmancuso: same thing here

src/masoniteorm/schema/platforms/SQLitePlatform.py
```diff
@@ -16,6 +16,7 @@ class SQLitePlatform(Platform):
         "small_integer": "SMALLINT",
         "medium_integer": "MEDIUMINT",
         "increments": "INTEGER PRIMARY KEY",
+        "uuid": "CHAR(36)",
```
josephmancuso: same thing here

src/masoniteorm/schema/Blueprint.py
```diff
@@ -85,8 +85,8 @@ def increments(self, column, nullable=False):
         )
         return self
 
-    def uuid_increments(self, column, nullable=False):
-        """Sets a column to be the UUID primary key representation for the table.
+    def uuid(self, column, nullable=False):
```
josephmancuso: You should accept a `length=36` here.

src/masoniteorm/schema/Blueprint.py
```diff
@@ -97,9 +97,7 @@ def uuid_increments(self, column, nullable=False):
         Returns:
             self
         """
-        self._last_column = self.table.add_column(
-            column, "uuid_increments", nullable=nullable
-        )
+        self._last_column = self.table.add_column(column, "uuid", nullable=nullable)
```
josephmancuso: add pass the `length=length` argument here

girardinsamuel:
Progress so far
- [x]  add "uuid" column in Blueprint, it's working (test_can_add_uuid_column)
- [ ]  handle auto generation of uuid when column is PK
- [ ]  add uuid version configuration

josephmancuso (changes requested):
This is perfect except you shouldn't hardcode the length in the platform. If you pass the length through the schema builder via `uuid('col', length=36)` it will compile down to `CHAR(36)`.

girardinsamuel:
Now the ideal would be to add `UUIDPrimaryKeyScope` to `_global_scopes` when a primary key constrained column is using uuid. Then the for loop on scopes already present on builder would be able to generate the uuid when inserting a new record.

But I don't know how/when to "link" `table.uuid().primary()` to `builder.set_global_scope(UUIDPrimaryKeyScope())` : @josephmancuso do you have an idea ? Am I on the right track 🤔 ?

src/masoniteorm/schema/Blueprint.py
josephmancuso: Uhh yeah looks right. I have to check the table tests

src/masoniteorm/query/QueryBuilder.py
```diff
@@ -349,6 +348,12 @@ def create(self, creates={}, query=False, id_key="id", **kwargs):
             creates = kwargs
 
         self.set_action("insert")
+
+        # here is an not a right place to do that but somewhere we should do this:
+        # TODO: if pk column is of type uuid, (how?)
+        # then add scope
+        self.set_global_scope(UUIDPrimaryKeyScope())
```
josephmancuso: This is done inside the mixin. The mixin has a boot where you call this method

src/masoniteorm/scopes/UUIDPrimaryKeyScope.py
```diff
@@ -0,0 +1,22 @@
+import uuid
+from .BaseScope import BaseScope
+
+
+class UUIDPrimaryKeyScope(BaseScope):
```
josephmancuso, line 5: It's a little confusing but now that we have a global scope we register it to a mixin and then the mixin gets inherited by the model

src/masoniteorm/scopes/UUIDPrimaryKeyScope.py
```diff
@@ -0,0 +1,22 @@
...
+            "_uuid", self.set_uuid_create, action="insert"
+        )
+
+    def on_remove(self, builder):
+        pass
+
+    def set_uuid_create(self, builder):
+        builder._creates.update(
+            {
+                # TODO: later how to fetch uuid version here ?
```
josephmancuso: Builder._model will give you the model. This is where you can check if it has any version attributes. You can set attributes on the mixin class which will get inherited down to the model when you inherit. You can default to 4

josephmancuso (changes requested):
This is perfect. The last missing piece of the puzzle is the Mixin class. You have your global scope created here. Now we need a mixin file like UUIDPrimaryKeyMixin or something. The mixin file will import this global scope you created here. 

When the model boots it will grab all parent classes that end with "Mixin" and call the boot method on it. In that boot method we call that builder.set_global_scope() method. 

I'll link an example. We do this same exact thing with time stamps.

josephmancuso:
@girardinsamuel check the TimeStampMixin class as an example

josephmancuso:
This mixin class is what the user will actually import and inherit to their models.

girardinsamuel:
> This is perfect. The last missing piece of the puzzle is the Mixin class. You have your global scope created here. Now we need a mixin file like UUIDPrimaryKeyMixin or something. The mixin file will import this global scope you created here.
> 
> When the model boots it will grab all parent classes that end with "Mixin" and call the boot method on it. In that boot method we call that builder.set_global_scope() method.
> 
> I'll link an example. We do this same exact thing with time stamps.

Thanks for the tips ! I was aware of the Mixin but was still looking forward to do it without it but I guess that's better with a Mixin rather than adding code where it should not belong. So we have a tradeoff between the two solutions 👍 
Now we can do:
```python
# migration
            table.uuid('id')
            table.primary('id')
            table.string('name')
            table.uuid('transaction_id')

# User.py
class User(Model, UUIDPrimaryKeyMixin):
    __uuid_version__ = 3
```

I am currently adding more tests on default global scopes, in this PR

josephmancuso (approved):
perfect. Solves the problem quite nicely.

Issue #179, opened 2020-10-02
__appends__ does not work on Collections


girardinsamuel:
This feature https://docs.masoniteproject.com/masonite-orm/models#adding-attributes
seems to only work on one model not a collection

# What I want to do
1. make `set_appends` work on collections
```python
# User model 
class User(Model):
    """User Model."""
   
    @property
    def name(self):
        return self.last_name + ' ' + self.first_name

# query
users = User.where("account_id", self.request.user().account.id).get()
users.set_appends(['name'])
```

2. Maybe add `__append__`  to model as in Orator, to automatically append properties defined in this array ?
```python
# User model 
class User(Model):
    """User Model."""
   __append__ = ["name"]

# query
User.all().serialize() # 'name' will be in the serialized data

PR #200, opened 2020-10-19
Fix/179: add set_appends on Collection

girardinsamuel:
Start implementing #179.

### DIFF
--- a/src/masoniteorm/collection/Collection.py
+++ b/src/masoniteorm/collection/Collection.py
@@ -8,6 +8,7 @@ class Collection:
 
     def __init__(self, items=[]):
         self._items = items
+        self.__appends__ = []
 
     def take(self, number: int):
         """Takes a specific number of results from the items.
@@ -291,6 +292,7 @@ class Collection:
 
     def serialize(self):
         def _serialize(item):
+            item.set_appends(self.__appends__)
             if hasattr(item, "serialize"):
                 return item.serialize()
             elif hasattr(item, "to_dict"):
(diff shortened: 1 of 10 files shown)


Marlysson:
@girardinsamuel This shouldn't made in Model class to use automatically the `__appends__` values when retrieved data by `.get` or `.all` method from querybuilder?

This already implemented in `.serialize()` method but works only when use the `set_append` method don't automatically ( better user experience )

josephmancuso:
@Marlysson @girardinsamuel thats a really good point. Maybe the set appends method just loops through all the models and calls set_appends on them. Then when it actually serializes, each model will have the set appends set on it already

girardinsamuel:
So this is the wanted behaviour:
* when `__appends__` is defined on a model, properties should be added **by default** to the serialized data of the instance and on the Collection
* when not defined, nothing is added
* If we want we can on the fly use `set_appends()` to add attributes to the serialized data of the instance and on the Collection

Is it correct ?

josephmancuso:
That's correct

src/masoniteorm/collection/Collection.py
```diff
@@ -291,6 +292,9 @@ def reverse(self):
 
     def serialize(self):
         def _serialize(item):
+            if self.__appends__:
+                item.set_appends(self.__appends__)
```
josephmancuso, line 296: I added this if check here because I'm thinking this will break when you go to serialize something that is not a collection of models. If you have a collections lists the list doesn't have set appends

PR #203, opened 2020-10-20
Feature/202 - Global connections

### LINKED ISSUE #202: Add a global transaction helper
I'm not even really sure how its done but somehow we need to do something like this:

```python
db.begin_transaction()

# do stuff here

db.rollback()
```

josephmancuso:
Closes #202 This is used to set global connections. This is useful when doing database transactions. With this PR, any connection class will actually fetch any globally set transactions when it makes a new connection.

Now we can do something like:

```python
db.begin_transaction() # sets a class as the global default
# operations that touch the default connection will all pass through the single connection object we set globally
db.commit() # commits the transactions and removes it from the global setting
db.rollback() # same as commit but rolls back the transaction
```

@girardinsamuel @Marlysson thoughts?


girardinsamuel:
It's great. Feel like this is the natural way to do it !
I guess now you "just have to add a decorator" which can use this global connection fetching.

src/masoniteorm/connections/ConnectionResolver.py
```diff
@@ -22,7 +23,68 @@ def set_connection_details(self, connection_details):
...
         from ..connections import ConnectionFactory
 
         ConnectionFactory.register(connection.name, connection)
+    
+    def begin_transaction(self, name=None):
+        from ..connections import ConnectionFactory
+        if name is None:
+            name = self.get_connection_details()['default']
+
+        connection = ConnectionFactory().make(name)(**self.get_connection_information(name)).make_connection().begin()
```
Marlysson: Great way to do that, very useful. What about you think of before create a new connection verify whether it's in `_connections`  ( and checking it's status..open, closed ) and return It? Already that all connections would be in this attribute.

PR #204, opened 2020-10-20
Fix/201: Add MySQL and Postgres database services for Github Actions tests

### LINKED ISSUE #201: Add different database services to Github Actions
I have seen that database credentials are added to Github Action, but no database service are defined so some tests are failing. 
It is possible to add database services in Github Actions to make MySQL and Postgres services available.

girardinsamuel:
Closes #201 Allow to run **all tests** of the codebase (excepted MSSQL related one) in CI.

@josephmancuso 
* there is one failing test, I don't really understand why there is this error.
* also I make few changes so that tests are working for CI and to be more generic, let me know if it's okay


josephmancuso:
@girardinsamuel the test is looking for the table_schema table. This test is just going a drop and add on the name column. 

looks like the connections are working perfectly but we need to somehow run migrations for the Postgres connection.

I suggest trying to make some migrations in the database/migrations directory maybe and then in the makefile run something like 

```
masonite-orm migrate --connection postgres
```

this will run the migrations based on the Postgres connection obviously and we will also need to set these Postgres database credentials up as environment configs and either use GitHub secrets to control it or just hard code it

girardinsamuel:
for CI we could hard code the parameters as done here, it's more straightforward and there is no security issue

girardinsamuel:
But I am wondering 🤔, I have seen tests succeeded before in Github Action on this repo but no services were installed ??

josephmancuso:
I was using heroku to host the free Postgres and mysql database lol. Really a short term solution. The GitHub action approach is better

girardinsamuel:
> @girardinsamuel the test is looking for the table_schema table. This test is just going a drop and add on the name column.
> 
> looks like the connections are working perfectly but we need to somehow run migrations for the Postgres connection.
> 
> I suggest trying to make some migrations in the database/migrations directory maybe and then in the makefile run something like
> 
> ```
> masonite-orm migrate --connection postgres
> ```
> 
> this will run the migrations based on the Postgres connection obviously and we will also need to set these Postgres database credentials up as environment configs and either use GitHub secrets to control it or just hard code it

I will try to do that, but for the database credentials it's already ok with PR setup (can switch to Github Secrets if you want but it's one step less to configure like this)

josephmancuso:
that's true. just have to keep in mind when running them locally on our machines it would be easier to just setup any credentials and to any table we need to for testing

girardinsamuel:
> that's true. just have to keep in mind when running them locally on our machines it would be easier to just setup any credentials and to any table we need to for testing

Sure, but's it's only hard-coded for CI, in local environment we have to define the variables in `.env.testing`.
But you're right, maybe we should put this clearly in "contributing doc" for this repo to help people running tests locally. In addition if you want to run tests you have to have both postgres and mysql services running and configured. So running tests is not as straightforward as for other packages.

girardinsamuel:
```
The 'None' connection does not exist
```
I also have this error in local, is the environment not loaded correctly ? 
```python
else:
            connection_details = connections.get(key) # key = 'postgres'
            connection = self._connections.get(connection_details.get("driver")) # self._connections = {} , this is the issue

        if connection:
            return connection

        raise Exception(
            "The '{connection}' connection does not exist".format(connection=connection)
        )
```

josephmancuso:
@girardinsamuel yeah I wonder If its the same issue with the entry point. I thought we fixed this but maybe not. I've been running migrate command through my masonite app so its loaded fine there because that database file gets imported pretty early in the Masonite boot process

girardinsamuel:
@josephmancuso yeah same for me.
I guess it's not the same issue because connection_details contains config data. It's _connections which is empty.
When instantiating `ConnectionResolver()` it should register all connections, but even if it's called at different places, here in the make() method of the ConnectionFactory, _connections is empty again.

girardinsamuel:
I even tried in `Migration.py`to swap 1 and 2 so that ConnectionResolver() call adds connections to the Factory, but it does not work:
```python
# 1
connection_class = ConnectionFactory().make(connection)
# 2
from config.database import ConnectionResolver
DATABASES = ConnectionResolver().get_connection_details()
```

ConnectionFactory could be a Singleton so that connections are registered ones and at successive calls we get connections.

josephmancuso:
@girardinsamuel hmm try doing this:

```python
# config/database.py
db = ConnectionResolver().set_connection_details(DATABASES)
# just set the current code to a variable

# migration.py
from config.database import db # get that variable from the config
DATABASES = db.get_connection_details()
```

girardinsamuel:
No it's not working either... I have debugged it step by step, but still does not understand why self._connections is reset to {} in ConnectionFactory...

girardinsamuel:
I tried to implement the ConnectionFactory as a singleton so after being instantiated once with some connections registered by the ConnectionResolver, when we try to fetch an instance of the ConnectionFactory, we still get the first one with the connections registered.
=>But I don't know why : the class is reinstantiated inside the MigrationCommand ... as if we were changing variables context inside Python process. So either with the original implementation or with the singleton it can't work

josephmancuso:
yeah something is a little funky with this connection class. it's not behaving like I've been expecting it to. It also works in one use case (via models and query builder and schema fetching connection details) but not others (like using it with the masonite-orm command ) even though the use of this class seems identical. I'll have to play with it tonight to see if we can structure it in a different way

josephmancuso:
@girardinsamuel fixing this branch now

josephmancuso:
@girardinsamuel THEY ALL PASS

josephmancuso:
Ok so I did a whole bunch of stuff in a few diff PR's. I feel like I broke everything and then fixed everything. 

So the biggest thing I did was that the connection factory class like you said kept getting reset somehow which still makes 0 sense because we were setting the class attribute and not the instance attribute so anytime you make a new instance it should inherit the class level attributes. Very weird. 

I think we could have gotten around it by hard coding the connections in the connection factory like before but I much rather register them because it makes the package much more pluggable and modular.

So I changed a bit on how we get the connection factory. One of the reasons we were having issues was because the connection factory kept on resetting with new instances we make. For example:

```python
from ..import ConnectionFactory

factory = ConnectionFactory().get_connection_details()
^^^^ something here was resetting.
```

So the obvious answer was to make sure we only ever have 1 instance. To do this I just set the connection factory on the ConnectionResolver class. If we set the connection resolver class to a variable:

```python
db = ConnectionResolver().set_connection_details(..)
```

we can then just import that. If we ever need the connection factory we can just always get the same one from the connection resolver:

```python
from config.database import db # == <masoniteorm.connections.ConnectionResolver object>

db.connection_factory.make(..)
```

girardinsamuel:
@josephmancuso really nice ! I had started yesterday to do it this way ! Glad this issue is resolved 👍

Issue #207, opened 2020-10-21
Be able to use .primary() as for unique() constraint


girardinsamuel:
Be able on a migration to use primary() this way:

```python
table.increments("user_id").primary()
```

josephmancuso:
@girardinsamuel increments is special. Increments assumes that its the primary key. 

increments is the same as "auto incrementing primary column" version of the table so its a bit of an all-in-one. 

What we really need to to be able to specify any one of the other column types as a primary key like:

```python
table.unsigned_integer('col').primary()
```

girardinsamuel:
@josephmancuso Yes that was a bad example but that's what I had in mind. 
And increments was confusing me I wasn't sure about what was ran under the hood. thanks for clarifying.

I will update the PR then

josephmancuso:
ok great just making sure :)

girardinsamuel:
I have updated the test in the PR, when I am using .primary() on a simple integer column

PR #208, opened 2020-10-21
Update primary constraint

girardinsamuel:
Implements #207


girardinsamuel:
@josephmancuso Is that okay for you ?

josephmancuso (approved):
perfect. that was easy :)

PR #210, opened 2020-10-22
Optional logging like Orator

Marlysson:
@josephmancuso

Optional logging by defining a key `log_queries` in connections informations in `DATABASES` dictionary in `database.py`

Currently there's the option of disable/enable logging during to the current connection:

```python
connection.enable_query_log() 
connection.disable_query_log()
```
But until now I didn't see some way to do like that ( the better way to this behavior )

```python
from config.database import db

db.disable_query_log() # get current connection and disable/enable
users = User.all() # query didn't logged
```

Initially I thought in override the connection informations in `ConnectionResolver` but it uses the `ConnectionFactory` that load again the `ConnectionResolver` ( db ) from `database.py` ( then don't make sense try override informations, already that will be replaced by a "new load" ( import ) ), but I see that this is the right way to disable the logging to current connection more implicit.


josephmancuso:
@Marlysson you're gonna need to set it on each connection class as well and inside the base connection there is a statement method. That is where the queries are logging currently. You'll need an if statement there to check whether to log or not

Marlysson:
Something diferent than what's defined in second commit?

josephmancuso (approved):

PR #213, opened 2020-10-22
Adding finds with fail option method on model.

### LINKED ISSUE #211: Add find_or_fail
Should be a new method on the model class. If should execute the normal find method but if there is no results it should execute a new `ModelNotFound` Exception

Marlysson:
@josephmancuso 

Whats the difference between `find_or_fail` and `first_or_fail`? Both uses the QueryBuilder.first() under the hood.

And the find_or_404 should have some http method import from somewhere?

- [x] find_or_fail
- [x] first_or_fail
- [x] find_or_404
- [x] Document each new method

Closes #211


josephmancuso:
They both use first but when you do find it adds a where based on the primary key. doing first_or_fail() will use query builders first but since there is no where clause it will literally just pull the first record in the table

josephmancuso:
So find or fail would run this query:

```
select * from users where id = 1 LIMIT 1
```

and first or fail will run this query:

```
select * from users LIMIT 1
```

josephmancuso:
first or fail would look like this:

```python
@classmethod
def first_or_fail(cls):
    result = cls().first()

    if not result:
        raise the exception here

    return result
```

Marlysson:
> They both use first but when you do find it adds a where based on the primary key. doing first_or_fail() will use query builders first but since there is no where clause it will literally just pull the first record in the table

first_or_fail should be added to `__passthrough__` ?

josephmancuso:
actually wouldn't be a bad idea. Thats good. I recommend putting these fail methods on the query builder class and then just adding them to the __pass_through__ attribute. This way the query builder also gets this feature

src/masoniteorm/query/QueryBuilder.py
```diff
@@ -1006,6 +1006,22 @@ def find_or_fail(self, record_id):
...
+
+        Returns:
+            Model|HTTP404
+        """
+
+
+        try:
+            return self.find_or_fail(record_id)
+        except ModelNotFound:
+            raise HTTP404()
```
Marlysson, line 1022: @josephmancuso I could repeat the whole code of find_of_fail method, but instead I do that....feeling that it doesn't smell good

src/masoniteorm/exceptions.py
```diff
@@ -4,3 +4,7 @@ class DriverNotFound(Exception):
 
 class ModelNotFound(Exception):
     pass
+
+
+class HTTP404(Exception):
```
Marlysson, line 9: @josephmancuso 
To do that create this HTTP exception level here. But smell not so good.

src/masoniteorm/query/QueryBuilder.py
josephmancuso: That's fine. I like this

src/masoniteorm/models/Model.py
```diff
@@ -86,6 +86,9 @@ class Model(TimeStampsMixin, metaclass=ModelMeta):
     __passthrough__ = [
         "all",
         "first",
+        "find_or_fail",
+        "first_or_fail",
+        "find_or_404",
```
josephmancuso, line 91: Perfect

src/masoniteorm/exceptions.py
josephmancuso: That's fine. I like it

josephmancuso (approved):

src/masoniteorm/query/QueryBuilder.py
```diff
@@ -987,6 +989,58 @@ def find(self, record_id):
 
         return self.where(self._model.get_primary_key(), record_id).first()
 
+    def find_or_fail(self, record_id):
+        """Finds a row by the primary key ID or raise an exception.
```
girardinsamuel, line 993: you should update the docstrings to differenciate all `find_or` methods 😉

src/masoniteorm/exceptions.py
```diff
@@ -1,2 +1,10 @@
 class DriverNotFound(Exception):
     pass
+
+
+class ModelNotFound(Exception):
+    pass
+
+
+class HTTP404(Exception):
```
girardinsamuel, line 9: Here we start to couple this independant package with http stuff... I really like this `find_or_404` method though (coming from Django !).
In addition, such an exception should be provided by Masonite core itself, isn't it ? 

But I don't really know how to deal (easily) with all that so I guess it could be let as is and improved later ?
@Marlysson @josephmancuso what do you think ?

girardinsamuel (changes requested):
Made a quick review ! Really cool features 👍

src/masoniteorm/exceptions.py
josephmancuso: for Masonite we have "ExceptionHandlers". We already use this feature for other things. We return different views based on the exceptions thrown. 

For example if you have ever used the `dd` helper, that throws a `DumpException`, we catch he exception and show a new view (that debug screen).

So it honestly doesn't matter what we name this exception, as long as its unique enough that frameworks are able to catch it

josephmancuso: So in this example we will need a new exception handler for HTTP404 in Masonite 3

girardinsamuel: Okay, makes sense.

Marlysson: Yeah, this make sense. So when use orm as standalone project should be this method unavailable?

josephmancuso: yeah it wouldn't make sense for someone to use this method if their framework didn't handle it. I'm 99% sure flask handles custom exceptions. I know I used exceptionite on a flask project which takes a similar approach to Masonite. 

So if someone used this with Flask they'd have to register the HTTP404 error somehow

josephmancuso (approved):

josephmancuso:
I added this to the documentation already. @Marlysson great job!

PR #215, opened 2020-10-23
Fix imports in migrations stubs

Marlysson:
When ran occours an exception because of old import.


josephmancuso (commented):
These are stub files. These are done when you make a migration in a normal project. so when you import to a normal project it would be masoniteorm.migrations import MIgration

PR #220, opened 2020-10-24
Preliminary implementation to show sql of migration selected.

### LINKED ISSUE #180: Add the ability to show the sql generated by a migration
It should be good to add a command which helps to show the SQL generated by a migration.
This feature is something special for Django. I like Laravel but I don't see anything like this when I need to show the SQL of a migration.

Example:
```bash
masonite-orm sqlmigration migration_for_users_table
``` 
Output:
```SQL
BEGIN;
CREATE TABLE users (
    id serial NOT NULL PRIMARY KEY,
   name varchar(150) NOT NULL
);
COMMIT;
```
You can check for Django using the link below
https://docs.djangoproject.com/en/3.1/ref/django-admin/#django-admin-sqlmigrate

Marlysson:
Options to show sql ( Closes #180 )
- [x] Migration selected
- [x] Connection used
- [x] Documentation
@josephmancuso @girardinsamuel

This is a introductory implementation to show sql. 

To catch the current blueprint of create or table I've need to assign it to a instance variable.


josephmancuso:
Taking a look now

josephmancuso:
Updated the PR to complete the feature.

This is now part of the normal migrate command. I removed the dedicated command.

We can use it like this:

```
python craft migrate -s
```

This will produce an output like this:

Screen Shot 2020-10-24 at 8 01 50 PM

josephmancuso:
not sure what the other observer files are in the diff. Those files are in 0.7 and this branch is up to date with 0.7 .. so can ignore those.

Marlysson:
So this is complete. Can we remove the commented code.
Make sense see individual migration sql?

josephmancuso:
I don't see the benefit of seeing an individual migration. You'll see it anyway in the list of unran migrations. we can wait to see if anybody requests seeing a single migration

Marlysson:
Ok

josephmancuso:
saw your checkbox. Also added this for rolling back. This was harder because dropping tables and sort of those one off sql statements don't use the blueprint class so had to do it a bit differently:

Screen Shot 2020-10-24 at 8 22 21 PM

josephmancuso:
@corentinalcoy how is this?

Marlysson:
Awesome!

girardinsamuel:
That's a very handy feature !

PR #221, opened 2020-10-24
Feature/219

### LINKED ISSUE #219: Add model events and observers
Orator and other ORMs like eloquent have this concept of model events and observers. We should be able to tie into events such as

Events should be as follows:

```
booting
booted
creating
created
deleting
deleted
hydrating
hydrated
saving
saved
updating
updated
```

And others that we need. 

We should be able to build out observers that can listen to an event such as `{model_name}.{action}`. In the observer we should also be able to listen for more global actions like `{model_name}.*` this would listen to all events of a specific model. and the opposite (`*.booting`) to listen to a specific event from all models or even `*.*` to listen to everything

josephmancuso:
Closes #219 

Implemented events:

- [x] creating
- [x] created
- [x] updating
- [x] updated
- [x] saving
- [x] saved
- [x] booting
- [x] booted  
- [x] deleting
- [x] deleted 
- [x] hydrating
- [x] hydrated
- [x] register observer to models 
- [x] Documentation


src/masoniteorm/models/Model.py
```diff
@@ -461,11 +463,22 @@ def save(self, query=False):
...
+
         if not query:
-            return builder.update(self.__dirty_attributes__)
+            result = builder.update(self.__dirty_attributes__)
+            self.observe_events(self, "saved")
+            return result
 
         return builder.update(self.__dirty_attributes__, dry=True).to_sql()
 
+    def observe_events(self, model, event):
```
Marlysson: What if remove the argument "model" from observers and uses "self": `self.__observers__` , `getattr(observer, event)`(self)

To be possible calling more cleaned:

```python
model.observe_events("event") 
```
Instead:

```python
model.observe_events(model,"event")
```

josephmancuso: great idea. Problem is that we need this same exact logic in 2 places. Some events happen only in the model (like hydrating) and some events happen in the query builder class (deleting, updating, etc). Because of this I abstracted the method out into a new class called `ObservesEvents` and inherited the model and query builder. To make it able to be used in both classes I need to accept a model

Marlysson: Great refactoring.

PR #226, opened 2020-10-25
Add missing paginator module

girardinsamuel:
paginator is missing from setup.py so installation fails


girardinsamuel:
@josephmancuso very small fix

PR #228, opened 2020-10-26
Add morphs helper to blueprint

girardinsamuel:
Added `morphs` to ease polymorphism relationships creation.
Still following Like,Article,Comment example we can now do:
```python
# likes migration
table.morphs("record")
```

- I did not setup an index automatically (between record_type and record_id) as done in Orator, should we do that ?
- I just created a test for proof, but what's the policy for testing "blueprints columns", they seems not tests except few, should I remove the test ? And should we do it for each connection ?

### DIFF
--- a/src/masoniteorm/schema/Blueprint.py
+++ b/src/masoniteorm/schema/Blueprint.py
@@ -352,6 +352,26 @@ class Blueprint:
         )
         return self
 
+    def morphs(self, column, nullable=False):
+        """Sets a column to be used in a polymorphic relationship.
+
+        Arguments:
+            column {string} -- The column name.
+
+        Keyword Arguments:
+            nullable {bool} -- Whether the column is nullable. (default: {False})
+
+        Returns:
+            self
+        """
+        self._last_column = self.table.add_column(
+            "{}_id".format(column), "unsigned_integer", nullable=nullable
+        )
+        self._last_column = self.table.add_column(
+            "{}_type".format(column), "string", nullable=nullable
+        )
+        return self
+
     def to_sql(self):
         """Compiles the blueprint class into a sql statement.
 
--- a/src/masoniteorm/schema/Blueprint.py
+++ b/src/masoniteorm/schema/Blueprint.py
@@ -352,7 +352,7 @@ class Blueprint:
         )
         return self
 
-    def morphs(self, column, nullable=False):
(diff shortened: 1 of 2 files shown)


girardinsamuel:
@josephmancuso ready for review with some questions

src/masoniteorm/schema/Blueprint.py
```diff
@@ -352,6 +352,26 @@ def unsigned_integer(self, column, nullable=False):
...
+
+        Keyword Arguments:
+            nullable {bool} -- Whether the column is nullable. (default: {False})
+
+        Returns:
+            self
+        """
+        self._last_column = self.table.add_column(
+            "{}_id".format(column), "unsigned_integer", nullable=nullable
+        )
```
josephmancuso: hmm. What about in the case someone does:

```
table.morphs('record').nullable()
```

? We might have to add self._last_column as a list or tuple and then nullable will act on all the elements in the list? Same actually goes for `table.timestamps()` too

josephmancuso (changes requested):
Looks great.

src/masoniteorm/schema/Blueprint.py
girardinsamuel: I did not thought about that, I'll look into it

girardinsamuel:
@josephmancuso and what about this part : 
> I did not setup an index automatically (between record_type and record_id) as done in Orator, should we do that ?

josephmancuso:
> I did not setup an index automatically (between record_type and record_id) as done in Orator, should we do that ?

I don't see why not. Maybe have that as a parameter. `morphs(indexes=True)`. have it default to true

src/masoniteorm/schema/Blueprint.py
```diff
@@ -399,12 +406,17 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
 
     def nullable(self):
-        """Sets the last column created as nullable
+        """Sets the last columns created as nullable
 
         Returns:
             self
         """
-        self._last_column.nullable()
+        _columns = []
```
girardinsamuel, line 414: @josephmancuso Here I deal with multiple columns locally in nullable() method but we could also have done it globally in the Blueprint class by  always using self._last_column as a list.
Don't know what's best

josephmancuso (approved):

PR #229, opened 2020-10-28
Feat/140: add transaction decorator

### LINKED ISSUE #140: Implements a decorator using a context to handle operations inside a transaction.
Create a decorator like ```@transaction``` to handle operation that should ran into a transaction.

Example:
```python
@transaction
def create_charge(self,**kwargs):
    # Doing some stuff
```

Behind the scenes the connection class should run:

```python
connection.begin()
connection.query()
connection.commit()
connection.rollback()
```

In just one step or rollback all case something gets wrong.

girardinsamuel:
It should be simple but I might have missed something, I have a weird error when running the test.

Closes #140


Marlysson:
Which error are you getting @girardinsamuel ?

girardinsamuel:
@josephmancuso 
```python
__________________ BaseTestQueryRelationships.test_transaction_decorator ___________________

self = <test_sqlite_transaction.BaseTestQueryRelationships testMethod=test_transaction_decorator>

    def test_transaction_decorator(self):
>       result = create_user()
E       TypeError: 'User' object is not callable

tests/sqlite/builder/test_sqlite_transaction.py:59: TypeError
```

I am not sure why it happens, I have looked it in details in the debugger

josephmancuso:
your decorator looks odd to me

girardinsamuel:
Actually I tried different syntaxes with function based decorators. But then I have seen that `@belongs_to` for example, is made with class based decorators, so I change my decorator to use class based version.

girardinsamuel:
I can rewrite with an other syntax if you want ? Or are you talking about the real content of it (the handling of error during transactions) ?

girardinsamuel:
This was the first decorator try :

```python
def transaction(*args, **kwargs):
    """Decorator to make atomic transactions."""
    def inner(func):
        connection_name = kwargs.get("connection", None)
        if not connection_name:
            # TODO: fetch default current connection
            connection_name = ""
        import pdb ; pdb.set_trace()
        db.begin_transaction(connection_name)
        try:
            # execute query
            func()
        except:
            import pdb; pdb.set_trace()
            db.rollback(connection_name)
        db.commit(connection_name)
    return inner
```

josephmancuso:
I don't really care how the decorator is structured lol. why are all the tests failing?

src/masoniteorm/connections/decorators.py.49ec9b0f8ff47a3ba27e2805c82b56ea.tmp
```diff
@@ -0,0 +1,20 @@
...
+        db.begin_transaction(self.connection)
+        try:
+            result = fn(*args, **kwargs)
+        except:
+            db.rollback(self.connection)
+            return None
+
+        # after function
+        db.commit(self.connection)
+        return result
```
josephmancuso: The reason you are getting the weird user error is because decorator are a pain in the ass to work with. Look closely what you're doing here. 

Because the result contains the user class, you are returning the user class here.

src/masoniteorm/connections/decorators.py.49ec9b0f8ff47a3ba27e2805c82b56ea.tmp
```diff
@@ -0,0 +1,20 @@
...
+class Transaction(object):
+
+    def __init__(self, connection="default"):
+        self.connection = connection
+
+    def __call__(self, fn, *args, **kwargs):
+        # before function
+        db.begin_transaction(self.connection)
+        try:
+            result = fn(*args, **kwargs)
```
josephmancuso: Here you are calling fn() as a function and passing args to it. The result of the create_user() function is the User class. So right now this result variable is a user class.

tests/sqlite/builder/test_sqlite_transaction.py
```diff
@@ -48,6 +54,10 @@ def test_transaction_globally(self):
         db.begin_transaction("sqlite")
         db.rollback("sqlite")
 
+    def test_transaction_decorator(self):
+        result = create_user()
```
josephmancuso: Now because of how python works the way you're using it, you are technically calling the user class here. (because you returned the user class in your decorator)

josephmancuso, line 58: Now because of how python works the way you're using it, you are technically calling the user class here.

josephmancuso (changes requested):

src/masoniteorm/connections/decorators.py
```diff
@@ -0,0 +1,20 @@
...
+class Transaction(object):
+
+    def __init__(self, connection="default"):
+        self.connection = connection
+
+    def __call__(self, fn, *args, **kwargs):
+        # before function
+        db.begin_transaction(self.connection)
+        try:
+            result = fn(*args, **kwargs)
```
josephmancuso: With that being said I think the solution is to just change this to:

```python
try:
    return fn
except:
    db.rollback(self.connection)
finally:
    db.commit(self.connection)
```

josephmancuso (changes requested):

src/masoniteorm/connections/decorators.py
josephmancuso: notice we're not calling fn anymore

girardinsamuel: Thank you I did not see this coming, should have paid more attention !

josephmancuso:
I tried testing this but it doesn't work at all ...

I think what you are trying to do is wrong here. The __call__ method gets called here when the decorator is set. you need logic to run when the underlying function is called

girardinsamuel:
Yes I know I just pushed some changes but I have seen that this does not work, I did not have time yet to fix it, I'll fix this by the end of the week if I can !
Thanks for the review 👍

josephmancuso:
@girardinsamuel ok no worries. No rush on this

src/masoniteorm/connections/decorators.py
```diff
@@ -1,17 +1,30 @@
...
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if exc_value:
             db.rollback(self.connection)
-        finally:
+        else:
             db.commit(self.connection)
+        return True
+
+
```
Marlysson, line 22: What is this "using" ? @girardinsamuel

girardinsamuel: It's nothing I forgot to remove that 😉

girardinsamuel:
@Marlysson @josephmancuso I am struggling a bit on this one ...

I got the class structure "working", ie I can use it as a decorator or as a context manager, but the logic of getting the default connection or the given connection and using it rollback or commit the transaction is wrong. I am not really familiar with this part so I can't figure out how to handle both "model transactions" and "transactions with query builder directly" in the same manner:

What we would like to do is (I guess) the following:
```python
# query builder style
with transaction() as db:
    db.table("users").create(...)

# model style
with transaction():
    User.create(...)
```

The `__enter__` method provide the `db` instance (right now it's working). But in the case of the "model style transactions" how to get the underlying query builder instance, and how to rollback transaction on the same instance than `User.create` used ?

Marlysson:
@girardinsamuel If I'm right the Model object has the attribute called `__resolved_connection__`  which returns the current connection. You can use it to create/open/close the current transaction.

```python
If there's a model, you get the current connection related to model
    db (ConnectionResolver)  from _model.__resolved_connection__
else:
    db (ConnectionResolver)  from config.database.db
```

girardinsamuel:
@Marlysson 
> ```python
> __resolved_connection__
> ```

Yes it looks like it's a way to get it but, the issue here is to be able to fetch `_model` or to be able to handle such a use case:
```
# model style
with transaction():
    user = User.create(...)
    profile = Profile.create(user=user, ...)
```
Can we, by design, have one connection used by both queries on the different models ? Because we need to be able to commit / rollback both transactions in this context.

Marlysson:
@girardinsamuel Wow, you're right! There's this case of different connections by operation. I'll look in deeper to help.

Marlysson:
@girardinsamuel @josephmancuso 
I was thinking here about this:

> Can we, by design, have one connection used by both queries on the different models ? Because we need to be able to commit / rollback both transactions in this context

A transaction starts with one specific connection, right? ( sqlite, default, staging... )

Is possible the model inside transaction uses other connection like?

```python
with transaction('sqlite'):
    User.on('postgres').create(...)
    Profile.on('mysql').create(...)
```

This seems very odd, start a transaction and run some operations in a new connection..  If not, we can avoid this use case of the scenario.

girardinsamuel:
Well to me it does not make sense to do that, I guess that is not a frequent use case !

Even with only one connection, I don't know how to fetch `__resolved_connection__` from the model used inside the context manager ?
For now I can only fetch a given connection by name or use the default if no connection provided, but I cannot fetch it from the model used because I can't fetch the model used inside the context manager.

josephmancuso:
> Can we, by design, have one connection used by both queries on the different models ? Because we need to be able to commit / rollback both transactions in this context

If you're going to do that you will need to start, commit or rollback both transactions of each database.

within this use case of the transactions you could do:

```python

class BankTransactions:

    @transaction('production') # production connection
    def add(self, amount):
        self.audit()
        # add amount

    @transaction('audit') # audit connection
    def audit(self):
        pass
```

girardinsamuel:
@josephmancuso yeah true. that's something we should take into account.

@Marlysson 
So quick summary for everyone of the issues I am facing: 
There are two major issues:
- the possibility of running transaction on different connections either with:
```python
class BankTransactions:

    @transaction('production') # production connection
    def add(self, amount):
        self.audit()
        # add amount

    @transaction('audit') # audit connection
    def audit(self):
        pass
```
or 
```python
with transaction():
    User.on('postgres').create(...)
    Profile.on('mysql').create(...)
```

- the other one is when using model style query and not direct builder queries, how can we fetch the underlying model's connection:
```python
with transaction('sqlite'):
    User.create(...)

# or through decorator

@transaction
def create():
    User.create(...)
```

girardinsamuel:
I summarised my thoughts and problems in the comments above. I am bit stuck on this right now, maybe because I don't have enough insights on the ORM internal operating.

@josephmancuso what are your thoughts on this ?

Thanks in advance

josephmancuso:
My thoughts are that decorators a pain in the a** to deal with for some reason and this is gonna require a bit more of a thought and this feature isn't really required as we have a way to deal with this already. An awesome feature for sure but not really a priority at the moment. I don't really want to spend time on getting this to work so I'm gonna close this for now. We can reopen if this becomes something requested for real use

girardinsamuel:
Yeah sure you're right ! We will try to crack it later ;)

PR #237, opened 2020-11-01
Add last() to query builder

girardinsamuel:
Right now we can do:
* `first()` and `last()` on Collections:
```python
User.all().first()
User.all().last()
```
* `first()` on the Model (pass through the query builder actually)
```python
User.first() # User with id 1
```
* but no `last()`to get the last row of the table. 
```python
User.last()
```

One way could be to do this by ordering by the primary ket in descending order. But what about running query not related to models or when `_model` is not defined in the query builder ?

@josephmancuso @Marlysson what do you think about this ?

### DIFF
--- a/src/masoniteorm/query/QueryBuilder.py
+++ b/src/masoniteorm/query/QueryBuilder.py
@@ -1007,18 +1007,18 @@ class QueryBuilder(ObservesEvents):
 
         return self.prepare_result(result)
 
-    def last(self, query=False):
+    def last(self, primary_key=None, query=False):
         """Gets the last record.
 
         Returns:
             dictionary -- Returns a dictionary of results.
         """
-
+        _primary_key = primary_key if primary_key else self._model.get_primary_key()
         if query:
-            return self.limit(1).order_by(self._model.get_primary_key(), direction="DESC")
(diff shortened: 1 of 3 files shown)


josephmancuso:
You can have last take the primary key. If None then take the primary key of the model. 

So like last(user_id)

src/masoniteorm/query/QueryBuilder.py
```diff
@@ -1007,6 +1007,22 @@ def first(self, query=False):
 
         return self.prepare_result(result)
 
+    def last(self, primary_key=None, query=False):
+        """Gets the last record.
+
+        Returns:
+            dictionary -- Returns a dictionary of results.
+        """
+        _primary_key = primary_key if primary_key else self._model.get_primary_key()
```
josephmancuso: I would actually change the naming of this to `column` because there will be times you want to get the last element based on a column. Like the highest price and do something like `last('price')`

josephmancuso (changes requested):

src/masoniteorm/query/QueryBuilder.py
girardinsamuel: Yes makes sense ! ✅

josephmancuso (approved):

PR #239, opened 2020-11-25
Feature/233

### LINKED ISSUE #233: Cannot eager load multiple eagers off of the same relation
[Issuehunt badges](https://issuehunt.io/r/MasoniteFramework/orm/issues/233)

Need to add eager loading support. This feature must have support for nested relationships to eager loading.

Currently there seems to be an issue with eager loading a relationship off the same eager loading relationship.

For example we can't do this:

```python
User.with_('articles.logo', 'articles.comments').get()
```

It doesn't seem to fetch properly. Doing a single eager load like `articles.comments.author` works fine but when we use eager load another nested relation from articles such as `articles.author`, it overwrite the original relationship.

The code is located in the QueryBuilder classes `prepare_result` method. This is where the eager loading is called and registered to the model

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/MasoniteFramework/orm/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

josephmancuso:
@Marlysson @girardinsamuel I think I've finally fucking figured this out. After like 2 weeks of procrastinating and trying not to work on this thing I think I got it!

Closes #233


girardinsamuel:
That's really great 💪  !!! I'll look into it to see how you solve this problem !

josephmancuso:
I did a little bit of a combination of both lists and dicts for relationships.

So for example if I had to eager load `User -> Profile` and I do `User.with_('profile')` then the eager loading would look something like:

```
[['profile']]
```

Its a list inside a list for reasons I'll explain in a bit.

The other scenario is when you have do a "nested eager load" (an eager load of a relationship):

In this scenario you will code something like `User.with_('profile.avatar')` which would now compile to something like:

```
[{"profile": ['avatar']}]
```

When It gets more complex it will add to this dict like so:

```
User.with_('profile.avatar', 'profile.logo')
#== [{"profile": ['avatar', 'logo']}]
```

Now the last scenario is when you have both nested and non nested. We need to account now for both lists and dicts:

```
User.with_("profile", ''profile.avatar', 'profile.logo')
#== [['profile'], {"profile": ['avatar', 'logo']}]
```

Since we needed to account for both the list and dict types it would look something like this now (notice its wrapped around a list which is where that first `[['profile']]` came from).

To Handle all this logic I just made a simple `EagerRelations()` class that is managed by the query builder class. and has a simple `register` method which determines how the eager and nested eager are handled.

Inside the query builder where I need to eager load this on the result, I know its nested if its a dict and I know its not nested if its a list.

josephmancuso:
that's the current solution. I'm not 100% sold on this solution and do think there is room for improvement somewhere but it's working right now and that's what we need. Improvements can be made a long the way

PR #244, opened 2020-11-28
Fix/241 - Fix nullable columns on schema alters

### LINKED ISSUE #241: Timestamp not setting nullable attribute in schema
When running a timestamp column like this:

```python
table.timestamp('due_date').nullable()
```

It results in this query:

```
ALTER TABLE `notes` ADD due_date TIMESTAMP 
```

This is wrong it should be 

```
ALTER TABLE `notes` ADD due_date TIMESTAMP NULL
```

josephmancuso:
Closes #241


josephmancuso:
Added fixes for mssql and sqlite and postgres as well

PR #249, opened 2020-11-29
Duplicated lines removed.

cinarizasyon:


josephmancuso (approved):
Looks great! Thanks!

josephmancuso:
@cinarizasyon if you could and if you have time, could you take a look at some of the other code? I wonder if this is the root cause of some of the weird issues I've had. Somehow I didn't know Python had this behavior.

cinarizasyon:
Of course, I can. I let you know if I find a similar place or another bug.

PR #250, opened 2020-11-29
Factory class refactored

cinarizasyon:
Hi Everyone,

dictionary parameter has a mutable default value. Let's change its default value is None and set the default value that we want manually in the methods.

In addition to the above, the definition and update processes of the called variable are not dependent on any variable in the conditions or loops. Therefore it would be good to move them to scopes of the methods.

### DIFF
--- a/src/masoniteorm/factories/Factory.py
+++ b/src/masoniteorm/factories/Factory.py
@@ -9,18 +9,18 @@ class Factory:
         self.model = model
         self.number = number
 
-    def make(self, dictionary=None, name="default"):
-        if dictionary is None:
-            dictionary = {}
+    def make(self, dictionary={}, name="default"):
+        dic = {}
+        dictionary.update(dic)
 
         called = self._factories[self.model][name](Faker())
-        called.update(dictionary)
+        called.update(dic)
 
-        if self.number == 1 and not isinstance(dictionary, list):
+        if self.number == 1 and not isinstance(dic, list):
             return self.model.hydrate(called)
-        elif isinstance(dictionary, list):
+        elif isinstance(dic, list):
             results = []
-            for index in range(0, len(dictionary)):
+            for index in range(0, len(dic)):
                 results.append(called)
             return self.model.hydrate(results)
         else:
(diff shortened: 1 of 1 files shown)


src/masoniteorm/factories/Factory.py
```diff
@@ -9,44 +9,43 @@ def __init__(self, model, number=1):
         self.model = model
         self.number = number
 
-    def make(self, dictionary={}, name="default"):
+    def make(self, dictionary=None, name="default"):
+        if dictionary is None:
+            dictionary = {}
```
josephmancuso, line 14: What is the reason for setting this to `None` here?

josephmancuso (changes requested):

src/masoniteorm/factories/Factory.py
cinarizasyon: dictionary argument had a mutable default value on the method signature. using an immutable default value is recommended if you want to set dictionary argument as {} when the method is called.

Also PyCharm warns in this case:

image

You can see that for more details: https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

josephmancuso: hmm I see. I'll have to test this out. I have no noticed this issue before in Python.

Wouldn't the solution instead of it defaulting to None wouldn't it be:

```python
def make(self, dictionary={}, name="default"):
        dic = {}
        dic.update(dictionary)
```

Setting it to None I feel is strange

cinarizasyon: your solution solves the issue, too. After your test, let's apply the solution that you feel comfortable to use it.

cinarizasyon: Is there any update? Which solution do you prefer?

josephmancuso: I think the example I put above where we keep everything a dictionary

cinarizasyon: I've just tried the above solution and tests failed. This solution forces the dic variable to be a dict in all cases however it may be a list in some cases. Whereas the dictionary variable must be an empty dict when only it is None according to the logic in the method. My solution provides this requirement. What do you think about how should we proceed?

josephmancuso: hmm that's true. I suppose your original solution with the None type is fine

cinarizasyon: Okay, I reverted the last 2 commits that cause tests were failed. Now, tests are passed.

josephmancuso (approved):

josephmancuso:
Thanks so much!

josephmancuso:
After some manual testing I realized this actually completely broke the factory class for any columns that are required to be unique. This change makes the factory class create the same dictionary of values where each one should really be a unique set of values lol

josephmancuso:
I now know why this was coded the way it was lol

PR #256, opened 2020-12-05
changed connection classes to use custom names

### LINKED ISSUE #255: Cannot name connection classes whatever we need.
Currently the connection class needs to be called either `mysql` or `postgres` and cannot be named something like `dev` and `staging`

josephmancuso:
Closes #255 

This PR fixes support having any connection names. Previously we were pulling the connection name and using the connection name as the key to fetch various registered connection details like the connection class, grammars and other things.

The correct way was to fetch the driver from the connection details and use that driver as the key to fetch connection details.


josephmancuso:
Tests are failing until I can work through all code

josephmancuso:
Do not merge. Requires refactoring cleanup

PR #272, opened 2020-12-17
Use case insensitive like for pgsql

### LINKED ISSUE #271: 'like' operator should default to 'ilike' for postgres
Thanks for fixing the like operator.

I just found out that Postgres has separate 'like' and 'ilike' operators (case-sensitive and case-insensitive respectively)

I would think that the grammar for 'like' and 'not like' in postgres default to 'ILIKE' and 'NOT ILIKE' as all other RDMS seem to have no distinction like this (no pun intended) and are all case-insensitive.

circulon:


josephmancuso:
@circulon These tests also need to be fixed. Same thing just changing LIKE to ILIKE

```
FAILED tests/sqlite/builder/test_sqlite_query_builder.py::SQLiteQueryBuilderTest::test_where_like
196
FAILED tests/sqlite/builder/test_sqlite_query_builder.py::SQLiteQueryBuilderTest::test_where_not_like
197
FAILED tests/sqlite/grammar/test_sqlite_select_grammar.py::TestSQLiteGrammar::test_where_like
198
FAILED tests/sqlite/grammar/test_sqlite_select_grammar.py::TestSQLiteGrammar::test_where_not_like
```

circulon:
@josephmancuso 

Oops totally updated the wrong grammar file.
Reverted the Sqlite grammar as there is no 'ILIKE' operater 
Added the postgres grammar change as I should have origanally.

josephmancuso (approved):

josephmancuso:
Looks Great!

PR #274, opened 2020-12-19
Patch/fix multiple order by

### LINKED ISSUE #275: Multiple order_by doesnt work
Currently using Model.order_by('col),order_by('other_col') throws a sql engine syntax error

Fix for this issue https://github.com/MasoniteFramework/orm/pull/274

circulon:
This Fixes using multiple .order_by() statements
Also there is a select query syntax fix that may have been an issue when using group by/having and order by in the same query

### DIFF
--- a/src/masoniteorm/query/grammars/BaseGrammar.py
+++ b/src/masoniteorm/query/grammars/BaseGrammar.py
@@ -263,7 +263,7 @@ class BaseGrammar:
                     column=self._table_column_string(column), direction=direction.upper()
                 )
             sql = self.order_by_string().format(
-                    order_columns=order_crit
+                order_columns=order_crit
             )
 
         return sql
--- a/src/masoniteorm/query/grammars/BaseGrammar.py
+++ b/src/masoniteorm/query/grammars/BaseGrammar.py
@@ -253,11 +253,15 @@ class BaseGrammar:
             self
         """
         sql = ""
-        for order_bys in self._order_by:
-            column, direction = order_bys
-            sql += self.order_by_string().format(
-                column=self._table_column_string(column), direction=direction.upper()
-            )
+        if self._order_by:
+            for order_bys in self._order_by:
+                if len(sql):
+                    sql += ", "
+                column, direction = order_bys
+                sql += self.order_by_string().format(
+                    column=self._table_column_string(column), direction=direction.upper()
+                )
+            sql = "ORDER BY " + sql
 
         return sql
 
--- a/src/masoniteorm/query/grammars/BaseGrammar.py
+++ b/src/masoniteorm/query/grammars/BaseGrammar.py
@@ -254,14 +254,17 @@ class BaseGrammar:
         """
         sql = ""
         if self._order_by:
+            order_crit = ""
             for order_bys in self._order_by:
-                if len(sql):
-                    sql += ", "
+                if len(order_crit):
+                    order_crit += ", "
                 column, direction = order_bys
-                sql += self.order_by_string().format(
+                order_crit += self.order_by_format().format(
                     column=self._table_column_string(column), direction=direction.upper()
                 )
-            sql = "ORDER BY " + sql
+            sql = self.order_by_string().format(
+                    order_columns=order_crit
+            )
 
         return sql
 
--- a/src/masoniteorm/query/grammars/MSSQLGrammar.py
+++ b/src/masoniteorm/query/grammars/MSSQLGrammar.py
@@ -30,7 +30,7 @@ class MSSQLGrammar(BaseGrammar):
     }
 
     def select_format(self):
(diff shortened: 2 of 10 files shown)


src/masoniteorm/query/grammars/BaseGrammar.py
```diff
@@ -253,11 +253,15 @@ def process_order_by(self):
...
-            )
+        if self._order_by:
+            for order_bys in self._order_by:
+                if len(sql):
+                    sql += ", "
+                column, direction = order_bys
+                sql += self.order_by_string().format(
+                    column=self._table_column_string(column), direction=direction.upper()
+                )
+            sql = "ORDER BY " + sql
```
josephmancuso: Grammar classes should abstract all SQL. Although all of the databases we currently support are the same, Databases we may support in the future may not. it really should be calling the child grammar classes to get this (`self.order_by_string()`)

src/masoniteorm/query/grammars/MSSQLGrammar.py
```diff
@@ -131,7 +131,7 @@ def table_string(self):
         return "[{prefix}{table}]"
 
     def order_by_string(self):
-        return "ORDER BY {column} {direction}"
```
josephmancuso, line 134: These need to go back

josephmancuso (changes requested):
Thanks for fixing! FYI all grammar classes should abstract SQL so we really can't have something like an `"ORDER BY"` in the `BaseGrammar` class. That text should really exist in the individual child grammar classes

src/masoniteorm/query/grammars/MSSQLGrammar.py
circulon: Hmmm 
Yeah I understand the Grammar needs to be abstracted.
Unfortunately if the order_by_string is reverted  you end up with a query that looks something like this in the current 0.9.3 version

SELECT ..... ORDER BY "name" DESCORDER BY "age" ASC .....  

Which is obviously broken and why I moved the prefix in the first portion of the string.


That said to keep the Grammar abstracted how about something like this in the BaseGrammar class...  a class var 

order_by_prefix = "ORDER BY"

This would allow easy future changes in any child class while still being able to be used in the process_order_by() like so

sql = self.order_by_prefix + sql

WDYT?

josephmancuso: Yeah I think that's fine. I don't really see what the difference is between prefixing it as a class property and putting it in these format string

We can do this: `sql = self.order_by_prefix + sql` and then maybe we can refactor later. I think the way I did some of the platform class coding is better for this situation anyway. Its a bit newer code and it handled this differently

circulon: To see the difference use the test for multiple order by against the current 0.9 version and it will fail and you will see the reason why the format string will not work when it is used in this manner. 

Anyway I will update this PR in the morning.
Cheers

circulon: @josephmancuso 
After thinking about this, looking at the existing format strings and keep the abstraction in place I have changed things slightly.

This adjustment now keeps formatting and keyword usage in their own class.
Let me know if there are  any pther considerations?

tests/postgres/grammar/test_select_grammar.py
```diff
@@ -69,6 +69,12 @@ def can_compile_with_order_by(self):
         """
         return """SELECT "users"."username" FROM "users" ORDER BY "users"."age" DESC"""
 
+    def can_compile_with_multiple_order_by(self):
+        """
+        self.builder.select('username').order_by('age', 'desc').order_by('name').to_sql()
+        """
+        return """SELECT "users"."username" FROM "users" ORDER BY "users"."age" DESC, "users"."name" ASC"""
```
josephmancuso, line 76: So this is a little tricky but how the select grammars tests work (and really how all test classes will work in the future when things aren't as fluid as they have been) Is that there is a `BaseTestCaseSelectGrammar` (see line 8) that has the test cases. those are inherited to the child test classes and call the related test methods to make the right assertions per grammar.

So for example a test case inside the `BaseTestCaseSelectGrammar` has a test case like `test_can_compile_with_multiple_order_by` which calls this method here (`can_compile_with_multiple_order_by` without the `test_` part). This allows the abstraction of tests that are the same to trickle down to the child test classes. So adding a test in 1 place basically calls tests for every grammar. Keeping it DRY

What I'm trying to say is that this test is not actually running without the `test_can_compile_with_multiple_order_by` inside the `BaseTestCaseSelectGrammar`

josephmancuso (changes requested):

circulon:
@josephmancuso 
Thanks for that explanation.

I havent done unit tests in python before so missed that kick off point.

Tests added for all grammars and now passing

josephmancuso (approved):
Looks good to me! 👍

josephmancuso:
Thanks!

PR #287, opened 2021-01-02
Created QueryLogger and changing the way to "turn on" logging in database.py

Marlysson:
@josephmancuso This is a almost done implementation. But I have one doubt:
How import defined `logger` from database.py in BaseConnection.py .... because database.py it's outside src package.


josephmancuso:
Can you explain a bit more on what this change is doing and how we would use this in a real app? And what benefit this has over what we have right now?

Marlysson:
@josephmancuso Sure! This way was a initial proposal. 

**Currently:**

We has that define all boilerplate using `logging` package: import handlers, define formatters, add handlers to root logger and after this it is done to import in every place of code. Then define a `self.log()` method in orm method to catch each hit in database.

**With changing:**

We just import the handler and define the default message ( this message is the dev that define ), there's other two internal formatter:

- query infos: query, binding and query_time
- system infos: time, module, thread, levelname, message...

This two messages is logged every time ( without the dev defined it ) plus the other where it define in `database.py`. Then we will have infos about query, system and user defined in every place logged.

**Use:**

*database.py*
```python
dev_defined_formatter = "It executed in {time}"
logger = QueryLogger(handlers, dev_defined_formatter)
```

*SomeController.py
```python
logger.log("My query", "SELECT ... ")
```

> 2006-02-08 22:20:02,165 **QUERY_INFO** My query query: "SELECT ..." bindings: ("1, "XPTO" ), query_time: "0.345ms" **SYSTEM_INFO** asctime module thread **USER DEFINED** It executed in 0.345ms

Maybe in log there a bit information? Yes. But it's all information available: queries logging, system logging and user. Maybe we could define a way to show less...

src/querylogger.py
```diff
@@ -0,0 +1,41 @@
+import logging
```
josephmancuso: class names like this should be `QueryLogger.py`

Marlysson: Solved. @josephmancuso There some way to import some attributes from databases.py into some files located in masoniteorm folder?

Marlysson:
@josephmancuso There some way to import some attributes from databases.py into some files located in masoniteorm folder?

josephmancuso:
I'm not 100% sold on this yet .. maybe it's because i don't understand it still but to me it looks like its just a wrapper around the `logging` module

josephmancuso:
@girardinsamuel what are your thoughts?

Marlysson:
@josephmancuso No problem. It's really just a wrapper around logging module. We could define messages in the old way too.

josephmancuso:
Yeah I'm not 100% sold on adding a whole new class to do the same thing we can do to the logging module. I think the idea Is good but in this context I don't see a benefit. I'll close this but I'm open to anyones thoughts on why it should be reopened

girardinsamuel:
> @girardinsamuel what are your thoughts?

Maybe this should go into an external package to add powerful logging capabilities into the ORM ? Because as josephmancuso said it's a wrapper around the exisiting logging module, I am not sure everyone will use it, and this is a bit opiniated to have those two log types at the same time.
But it could be very useful for debugging !!

Issue #276, opened 2020-12-21
Add new builder method scopes for soft deleting 


josephmancuso:
Need to add the following scopes or methods to the `SoftDeletesMixin` class:

* `only_trashed()`
* `restore()`
* `force_delete`

girardinsamuel:
Yeah that's exactly what I was looking for the pingcrm inertia demo ! 

I am struggling to get only trashed users as we should check that users have a `deleted_at` field not null or not empty.. => `where_not_null("deleted_at")`  does not seem to work, I guess that the default value for `deleted_at` is not the SQL `NULL` but maybe "" ?

girardinsamuel:
I will open a PR

josephmancuso:
This is done

PR #290, opened 2021-01-03
Improve soft delete mixin

girardinsamuel:
Implements #276 
- [x] only_trashed()
- [x] restore()
- [x] force_delete

### DIFF
--- a/src/masoniteorm/scopes/SoftDeleteScope.py
+++ b/src/masoniteorm/scopes/SoftDeleteScope.py
@@ -9,10 +9,8 @@ class SoftDeleteScope(BaseScope):
         builder.set_global_scope(
             "_query_set_null_on_delete", self._query_set_null_on_delete, action="delete"
         )
-        builder.set_global_scope(
-            "only_trashed", self._only_trashed, action="select"
-        )
         builder.macro("with_trashed", self._with_trashed)
+        builder.macro("only_trashed", self._only_trashed)
 
     def on_remove(self, builder):
         builder.remove_global_scope("_where_null", action="select")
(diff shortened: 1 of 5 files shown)


josephmancuso:
the restore method should just turn the query into an update method:

should run this query:

```python
builder.update({"deleted_at": None})
```

Force delete should just delete the record using a DELETE query

josephmancuso:
If you pull in 0.9 it will update some of those broken tests

josephmancuso:
Changed this to be based on new 1.0 branch

josephmancuso:
Added force delete

josephmancuso:
Added restore method

josephmancuso (approved):
Looks like I finished the last 2 methods required to close this ticket out! Great job!

src/masoniteorm/scopes/SoftDeleteScope.py
```diff
@@ -21,6 +25,16 @@ def _with_trashed(self, model, builder):
         builder.remove_global_scope("_where_null", action="select")
         return builder
 
+    def _only_trashed(self, model, builder):
+        builder.remove_global_scope("_where_null", action="select")
+        return builder.where_not_null("deleted_at")
+
+    def _force_delete(self, model, builder):
+        return builder.remove_global_scope(self).set_action("delete")
```
girardinsamuel, line 33: @josephmancuso  I am just not sure to understand why we should call `remove_global_scope(self)` before setting action type ? it is to avoid other scope to apply so that `force_delete` really deletes ? If this was not there, the force_delete would be modified to an update of deleted_at with a timestamp through `_query_set_null_on_delete` ?

josephmancuso: >  I am just not sure to understand why we should call remove_global_scope(self) before setting action type

doesn't matter which order we do it in. 

```python
builder.set_action("delete").remove_global_scope(self)
```

Works just as well. We do have to remove the scopes because yeah we don't want other scopes changing the action type back to update. We might not even have to set the action here, just removing the scope might actually work ..

PR #305, opened 2021-01-09
Feat/289: Add __visible__ to model attributes for serialization

### LINKED ISSUE #289: Add __visible__ to model serialization
Right now, when serializing a model it will put all model fields + properties (what has been defined in __append__).

I have seen that we can start from a dict given as argument of the `serialize()` method, but I have not really see a way to define how the model should be serialized without using Masonite API.

Here I just want - on the fly - to serialize e.g. the User model with three attributes: id, name, email. `name` is a property.
I feel we would like to do something like: 
```
User.find(1).serialize(["id", "name", "email"]) # {"id": 1, "name": "John Doe", "email": johndoe@example.com}
```
OR we could also add something in the model such as : `__serialized_fields__ = ["id", "name", "email"]` 

If you confirm it's not possible right now and this implementation is correct, I'll open a PR.
We should also update documentation to explain model serialization I have not seen it in orm docs.

girardinsamuel:
Closes #289.

I also fixed tests running with mysql for the model

### DIFF
--- a/tests/mysql/model/test_model.py
+++ b/tests/mysql/model/test_model.py
@@ -156,20 +156,6 @@ if os.getenv("RUN_MYSQL_DATABASE", False) == "True":
                 {"name": "Joe", "email": "joe@masonite.com"}, profile.serialize()
             )
 
-        def test_serialize_with_visible_and_initial_data(self):
-            profile = ProfileSerializeWithVisible.hydrate(
-                {
-                    "name": "Joe",
-                    "id": 1,
-                    "password": "secret",
-                    "email": "joe@masonite.com",
-                }
-            )
-            self.assertTrue(
-                {"name": "Joe", "email": "joe@masonite.com", "role": "admin"},
-                profile.serialize({"role": "admin"}),
-            )
-
         def test_serialize_with_visible_and_hidden_raise_error(self):
             profile = ProfileSerializeWithVisibleAndHidden.hydrate(
                 {
--- a/tests/mysql/model/test_model.py
+++ b/tests/mysql/model/test_model.py
@@ -1,4 +1,5 @@
 import datetime
+from inspect import Attribute
 import json
 import os
 import unittest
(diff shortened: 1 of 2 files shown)


girardinsamuel:
@josephmancuso I made a new failing test to explain my misunderstanding about the initial dict we can provide to `serialize()`method of the model.
It's documented that it's data to start from so we might think that it will start from this dict and other model attributes will added (whatever choices are made with __hidden__ and __visible__) => but actually it skip model attributes to use only this initial dict. 
I was expecting the first behaviour (as put in my failing test `test_serialize_with_visible_and_initial_data`).

Maybe I am missing the use case for the actual behaviour ? 

The PR is ready, I just have to the remove the failing test after we discussed a bit about that 👍

josephmancuso:
@girardinsamuel I'm not even 100% sure we are even using that anymore. It might have been left there when we were trying to figure out the relationship issues. I don't even see any tests for it. I can't think of a valid use case for it and I just removed it and ran the tests and everything passed soo. not sure why we should keep it. It's deff not a requirement to keep it for setting a dictionary to start from. I can't think of a reason to use that in an actual app

girardinsamuel:
Totally agree !!

tests/mysql/model/test_model.py
```diff
@@ -126,6 +143,45 @@ def test_serialize_with_hidden(self):
...
+                    "id": 1,
+                    "password": "secret",
+                    "email": "joe@masonite.com",
+                }
+            )
+            self.assertTrue(
+                {"name": "Joe", "email": "joe@masonite.com"}, profile.serialize()
+            )
+
+        def test_serialize_with_visible_and_initial_data(self):
```
josephmancuso: so we can just remove this test and remove that serialized_dictionary in the parameter list and set serialized_dictionary to be the model attributes

josephmancuso (changes requested):

girardinsamuel:
@josephmancuso i made the changes !

josephmancuso (approved):

josephmancuso:
FYI your original description said "implements #289". This links the issue in the issue screen but you should use a linking keyword for actions. For example if you did "Closes #289" then this PR will automatically close when merged

https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #306, opened 2021-01-10
Feature/296

### LINKED ISSUE #296: Add "joins" method to query builder
Currently we have a way to join using the table name and the key relationships.

Something like this:

```python
Article.join('comments', 'articles.id', '=', 'comments.article_id').join('users', 'articles.user_id', '=', 'users.id')
```

This is a bit cumbersome so we need to add a new method that fetches this information via the relationships defined on the model. So something like this:

```python
Article.joins('comments', 'users')
```

What this should do is loop over each relationship in the parameter list and fetch the relationships keys from them and build out the joins.

It should be really simple because the joining method already works we just need to build the method using the model relationships

josephmancuso:
Closes #296


josephmancuso:
@GeorgeJoLo the solution is here if you want to see it

GeorgeJoLo:
Of course, I'm going to give a look at this!
Thank you!

PR #318, opened 2021-01-21
add TIME type

erhuabushuo:


src/masoniteorm/schema/Blueprint.py
```diff
@@ -186,6 +186,21 @@ def date(self, column, nullable=False):
...
+        Arguments:
+            column {string} -- The column name.
+
+        Keyword Arguments:
+            nullable {bool} -- Whether the column is nullable. (default: {False})
+
+        Returns:
+            self
+        """
+        self._last_column = self.table.add_column(column, "date", nullable=nullable)
```
Marlysson: @erhuabushuo Is it here "date" type really?

PR #320, opened 2021-01-21
add TIME type

erhuabushuo:
* add TIME type
* add native jsonb type to postgres


Marlysson:
@erhuabushuo Welcome to the Masonite project. 
You can open the PR and send all of you changes without close it to add new ones . Just send a couple of commits with your changes even if it's to fix from previous commit.

PR #321, opened 2021-01-21
add missed data type

erhuabushuo:
Closes #320


josephmancuso (changes requested):
Everything looks good. Does any changes to the MYSQL grammar need to be done?

erhuabushuo:
@josephmancuso MySQL grammar doesn't need to be change,  other platform integer types do not support length, so they need to be define at types_without_lengths.

josephmancuso (approved):

josephmancuso:
This is all great! Thanks! Great Job!

PR #325, opened 2021-01-23
Feature/322

### LINKED ISSUE #322: Add Pivot Table support (belongs_to_many)
I see that in Orator there is a  facility for working with Pivot Tables
https://orator-orm.com/docs/0.9/orm.html#working-with-pivot-table

Is this a feature of Masonite Orm using the current decorators? or maybe @morph_to (which i'm not sure I understand anyway)

My investigations have only pointed out that I will have a fun time making the traversal work.

Any pointers/snippets  would be greatly appreciated

josephmancuso:
Closes #322 

This adds a few features.

The first is being able to specify a callable in the where_in method to specify a sub query:

```python
builder.where_in('id', lambda q: q.select('column').where('active', 1))
```

The second feature is adds a belongs to many relationship to get the related rows through a pivot table.

There are 2 ways to do this: a join or a subquery. I think a join is faster for small data sets but a subquery should be faster with large data sets. I think and in theory. Need to check this.

This is the example query that should generate

```sql
-- query 1 170us
SELECT 
    "products".*, 
    "product_store"."store_id", 
    "product_store"."product_id", 
    "product_store"."updated_at" AS m_reserved_1, 
    "product_store"."created_at" AS m_reserved_2 
    "product_store"."id" AS m_reserved_3
FROM "stores" 
INNER JOIN "product_store" 
    ON "product_store"."store_id" = "stores"."id" 
INNER JOIN "products" 
    ON "product_store"."product_id" = "products"."id"
WHERE "stores"."id" = 1
```

### TODO:

- [x] Add `pivot` attribute to the model from the relationship
- [x] Specify the pivot table name
- [x] Pivot Table attribute alias
- [x] Add more tests for testing query result
- [x] Add Eager Loading support


josephmancuso:
@girardinsamuel @Marlysson 

This also adds 3 more features.

The first feature is probably more of a fix but being able to specify a column in a normal select statement:

```python
.select("table.column")
```

and it compiles correctly to: `"table"."column"`

The other is adds the ability to specify an asterisk instead of a column and it compiles correctly:

```python
.select("table.*")
```

and it compiles correctly to: `"table".*`

Lastly it adds the ability to specify a column alias directly in a select method:

```python
.select("table.username as name")
```

and it compiles correctly to: `"table"."username" AS name`

src/masoniteorm/relationships/BelongsToMany.py
```diff
@@ -0,0 +1,155 @@
...
+            }
+
+            if self.with_timestamps:
+                pivot_data.update(
+                    {
+                        "updated_at": getattr(p, "m_reserved_1"),
+                        "created_at": getattr(p, "m_reserved_2"),
+                    }
+                )
+            setattr(p, self._as, Pivot.hydrate(pivot_data))
```
josephmancuso: Need to set the table and the connection here

src/masoniteorm/relationships/BelongsToMany.py
```diff
@@ -0,0 +1,155 @@
...
+
+        return result
+
+    def table(self, table):
+        self._table = table
+        return self
+
+    def get_related(self, query, relation, eagers=None):
```
josephmancuso, line 134: Need to retest the eager loading

tests/sqlite/relationships/test_sqlite_relationships.py
```diff
@@ -49,6 +49,28 @@ def get_is_admin(self):
...
 
+class Store(Model):
+
+    __connection__ = "dev"
+
+    @belongs_to_many("store_id", "product_id", "id", "id")
+    def products(self):
+        return Product
+
+    @belongs_to_many("store_id", "product_id", "id", "id", table="product_table")
```
girardinsamuel, line 60: @josephmancuso I am not sure to understand why this `product_table` should be specified and also the way it is specified (through a relationship and not just an attribute) 🤔

PR #326, opened 2021-01-24
Feature/240

### LINKED ISSUE #240: Add list_table_columns to schema builder
needs a way to list the table columns so we can rework the model:doctstring command.

In order it was called `list_table_columns`:

example in orator:

```python
            conn = (
                DB.connection(self.option("connection"))
                .get_schema_manager()
                .list_table_columns(self.argument("table"))
            )
```

Marlysson:
Closes #240 @josephmancuso I just dont know how testing this command. I've use the cc.py file and everything works fine to postgres and sqlite with all types tested.


josephmancuso:
@Marlysson fixed a small issue where its possible a column match is not found. 

Something like this really doesn't need to be tested. This doesn't affect major functionality of the ORM and its difficult enough to test commands.

I did test this with other databases like mysql and postgres and it doesn't look like there is any work on this? Does this PR only add support for model docstrings for SQLIte? Or are you able to / plan on adding more

Marlysson:
Ok, I'll remove the tests.

The blueprint on issue related to this PR for PostgreSQL show the docstrings.

But really there some fields that are missing in mapped types cause isn't yet in mapped types, like timestamp with timezone... I'll see all of our types in postgres to get done.

But which blueprint do you tested with postgres that dont work?

I plan finish sqlite and postgres and after mysql and mssql ( this two I'm not have them installed yet and low knowledge about its grammar )

josephmancuso:
I didn't test Postgres. I just tested SQLite and MySQL. SQLite looked good and then MySQL returned a blank doc string but I think that's because there's no work done on MySQL

girardinsamuel:
@Marlysson FYI if you want to test commands, I have done it on masonite repo, it's not very difficult. Here is an example https://github.com/MasoniteFramework/masonite/blob/3.0/tests/commands/test_new.py (where I am testing craft new)

- your should register the command you want to test in setUp (L22 > L25)
- then in your test call it (L40)
- and assert what has been printed out (L41 check for error, L84 check for success)

Marlysson:
@josephmancuso I think we have a "problems" here in sqliteplatform.

1. These types:
```python
blueprint.big_integer("b")
blueprint.big_increments("c")
```

```python
{
    "big_integer": "BIGINT",
    "big_increments": "BIGINT"
}
```

All mapping to "BIGINT" in database. How can we get the correct masonite-orm type of Blueprint?

2. Increments type ( as primary key )
```python
{ 
    "increments": "INTEGER PRIMARY KEY",
}
```
When persists in database it's transformed to just "INTEGER", Can I transform the mapping to just "INTEGER" ? And so, it comes to same problem with 1... two masonite-orm types with same database types. How get correct by value in reversed_map.

josephmancuso:
@Marlysson You are right. I think big_increments should be `BIGSERIAL`

Marlysson:
With some of this changes ( mapping types ) some tests in sqlite(grammar/platform) will break. It's ok?

josephmancuso:
@Marlysson depends on which tests are breaking and how they are breaking. Usually a broken test means that the public facing API of the package has changes which can't happen at this point.

josephmancuso:
or are you saying changing to BIGSERIAL need to change tests? If that's the case then that's fine. If by fixing tests we are fixing issues that's fine. 

Another good rule of thumb is that if we have to change the documentation then it is breaking and we can't add that in version 1

Marlysson:
> or are you saying changing to BIGSERIAL need to change tests? If that's the case then that's fine. If by fixing tests we are fixing issues that's fine.

Yeah, exactly that. Break because the sql are matching with specific sql and change the type will change the sql.

josephmancuso:
that's fine then. that's more of a bug fix than anything else

Marlysson:
@josephmancuso Done for all of of sqlite types. I will keep this as done for sqlite first, and after your revision, close the PR and will open a new PR for each platforms to don't keep the PR so many commits.

Made some changes in sqlite types to match in a better way with some types.

- Some changes was to better match with real types and for keep ( as far as possible ) different types to get the reverse relation between blueprint method and batabase types.
- New field type "CLOB" it's here https://www.sqlite.org/datatype3.html for references.

girardinsamuel:
@Marlysson what is the progress on this ? Can I help you ? 😉

Marlysson:
@girardinsamuel It's more laborious than complicated. But sure, you can.

When I fetch some fields/columns it returns same values from sabe types, difficulting the detecting of correct column type.

I will build the suite test that this happen and show you.

PR #331, opened 2021-01-25
Add random to collection

girardinsamuel:
```
User.all().random() # will return a random user or nothing if no records
```

### DIFF
--- a/src/masoniteorm/collection/Collection.py
+++ b/src/masoniteorm/collection/Collection.py
@@ -277,7 +277,9 @@ class Collection:
 
     def random(self):
         """Returns a random item of the collection."""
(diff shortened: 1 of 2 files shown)


josephmancuso (approved):

josephmancuso:
Awesome! Just needs documentation.

girardinsamuel:
@josephmancuso sorry for that, I changed my mind and updated random function to be able to provide a count to the random function (and it will return a Collection in this case) to be able to continue chaining methods.

I will add doc for multiple collections methods at the same time as they are not documented

girardinsamuel:
I finally opened a new PR for the update
https://github.com/MasoniteFramework/orm/pull/333

PR #333, opened 2021-01-25
Improve random collection method

girardinsamuel:
I'll document this in the doc but quickly
```
users.random() # == one user instance
users.random(3) #== a collection of 3 users picked randomly
users.random(1) #== a collection of 1 user picked randomly
```
When every you give a count to random it will return a Collection (so it can be chained) else it will just return the element.
Edges cases are handled (count > collection length or empty collection)

### DIFF
--- a/src/masoniteorm/collection/Collection.py
+++ b/src/masoniteorm/collection/Collection.py
@@ -275,12 +275,18 @@ class Collection:
         self[key] = value
         return self
 
-    def random(self):
+    def random(self, count=None):
         """Returns a random item of the collection."""
-        if self.count() == 0:
+        collection_count = self.count()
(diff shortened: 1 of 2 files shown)


src/masoniteorm/collection/Collection.py
```diff
@@ -275,12 +275,18 @@ def put(self, key, value):
         self[key] = value
         return self
 
-    def random(self):
+    def random(self, count=None):
```
girardinsamuel, line 278: Did not want to define count=1 by default, as I would like the behaviour described in the PR:
- count is not specified  => returns one element
- count is specified => returns a collection (event if 1)

josephmancuso (approved):

PR #349, opened 2021-01-30
added append passthough

### LINKED ISSUE #330: Querying trashed model does not work
On a soft deletable User model I get this:

```python
user = User.with_trashed().find(self.request.param("user"))
# == class model 'User' has no attribute with_trashed
```
I observe also this error on the line shown in [ORM documentation](https://orm.masoniteproject.com/models#soft-deleting):
```
>>> from app.User import User
>>> User.with_trashed().all()
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/Users/samuel/.pyenv/versions/pingcrm_masonite/lib/python3.9/site-packages/masoniteorm/models/Model.py", line 40, in __getattr__
    return getattr(instantiated, attribute)
  File "/Users/samuel/.pyenv/versions/pingcrm_masonite/lib/python3.9/site-packages/masoniteorm/models/Model.py", line 510, in __getattr__
    raise AttributeError(f"class model '{name}' has no attribute {attribute}")
AttributeError: class model 'User' has no attribute with_trashed
```

But the soft deletes mixin is working otherwise. For example I can do this:
```
User.where("id", 4).count() #== 0
User.where("id", 4).with_trashed().count() # == 1
```

Here is my model: https://github.com/girardinsamuel/pingcrm-masonite/blob/fix/3.0/app/User.py

josephmancuso:
Closes #330

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -204,6 +204,12 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
             self._booted = True
             self.observe_events(self, "booted")
 
+            self.append_passthrough(list(self.builder._macros.keys()))
+
+    def append_passthrough(self, passthrough):
+        self.__passthrough__ += passthrough
+        return self
+
     @classmethod
     def get_table_name(cls):
         """Gets the table name.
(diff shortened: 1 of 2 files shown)


josephmancuso:
@girardinsamuel simple fix here actually. Just have to take the macros from the query builder when the model boots and append it to the pass through

PR #354, opened 2021-01-31
Add deleted_at attribute setting

### LINKED ISSUE #297: Add a way to set the deleted_at column used for soft delete scopes
Right now the `deleted_at` column is hard coded, There should be a way to override this column name.

First step is adding a new attribute to this file `src/masoniteorm/scopes/SoftDeletesMixin.py`. The attribute should be called `__deleted_at__` along with a method to fetch this attribute `get_deleted_at_column()`:

```python
class SoftDeletesMixin:
    """Global scope class to add soft deleting to models."""

    __deleted_at__ = "deleted_at"

    def boot_SoftDeletesMixin(self, builder):
        builder.set_global_scope(SoftDeleteScope())

    def get_deleted_at_column(self):
        return self.__deleted_at__
```

Then the next set of changes are in this file: `src/masoniteorm/scopes/SoftDeleteScope.py`.

Any place you see `"deleted_at"`, it should be fetched like this:

```python
return builder.remove_global_scope(self).update({builder._model.get_deleted_at_column(): None})
```

girardinsamuel:
Fixes #297 

Well actually I tried it out and it's not that simple (at least for me)
The tests for soft delete are done at query builder level, without model involved, so `builder._model` is None and we cannot fetch the name of the deleted_at column. => The only place where it has been defined is when calling `soft_deletes()` in the migration file (where the default value "deleted_at" can be overriden).

```python
def _only_trashed(self, model, builder):
    builder.remove_global_scope("_where_null", action="select")
    # => builder._model is None here when using query builder and not model
    return builder.where_not_null(builder._model.get_deleted_at_column())
```

So I am not sure how to proceed then, we should maybe use a switch to fetch the column name depending if the model is defined 🤔 ? Does not look very solid to me.

@Marlysson @josephmancuso some insights ?

### DIFF
--- a/src/masoniteorm/scopes/SoftDeleteScope.py
+++ b/src/masoniteorm/scopes/SoftDeleteScope.py
@@ -4,6 +4,9 @@ from .BaseScope import BaseScope
 class SoftDeleteScope(BaseScope):
     """Global scope class to add soft deleting to models."""
 
+    def __init__(self, deleted_at_column="deleted_at"):
+        self.deleted_at_column = deleted_at_column
+
     def on_boot(self, builder):
         builder.set_global_scope("_where_null", self._where_null, action="select")
         builder.set_global_scope(
@@ -19,7 +22,7 @@ class SoftDeleteScope(BaseScope):
         builder.remove_global_scope("_query_set_null_on_delete", action="delete")
 
     def _where_null(self, builder):
-        return builder.where_null(builder._model.get_deleted_at_column())
+        return builder.where_null(self.deleted_at_column)
 
     def _with_trashed(self, model, builder):
         builder.remove_global_scope("_where_null", action="select")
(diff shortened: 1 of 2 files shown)


src/masoniteorm/scopes/SoftDeletesMixin.py
```diff
@@ -4,5 +4,10 @@
 class SoftDeletesMixin:
     """Global scope class to add soft deleting to models."""
 
+    __deleted_at__ = "deleted_at"
+
     def boot_SoftDeletesMixin(self, builder):
         builder.set_global_scope(SoftDeleteScope())
```
josephmancuso: We can pass it here?
```suggestion
        builder.set_global_scope(SoftDeleteScope(column=self.__deleted_at__))
```

?

src/masoniteorm/scopes/SoftDeleteScope.py
```diff
@@ -19,23 +19,23 @@ def on_remove(self, builder):
         builder.remove_global_scope("_query_set_null_on_delete", action="delete")
 
     def _where_null(self, builder):
-        return builder.where_null("deleted_at")
+        return builder.where_null(builder._model.get_deleted_at_column())
```
josephmancuso: then inside here we can just do self.deleted_at_column

src/masoniteorm/scopes/SoftDeletesMixin.py
```diff
@@ -4,5 +4,10 @@
 class SoftDeletesMixin:
     """Global scope class to add soft deleting to models."""
 
+    __deleted_at__ = "deleted_at"
```
josephmancuso, line 7: This value gets overidden at the model level. Since this Mixin class inherited by the model. We can just pass this value into the scope class. Inside the constructor of the scope we can set it on the scope class itself.

```python
class SoftDeleteScope:

    def __init__(self, deleted_at_column="deleted_at"):
        self.deleted_at_column = deleted_at_column
```

josephmancuso (changes requested):
This will work right?

josephmancuso:
@girardinsamuel 

> we should maybe use a switch to fetch the column name depending if the model is defined

No. switches are almost never a good idea ever

girardinsamuel:
> No. switches are almost never a good idea ever

I agree, wasn't really happy about this. I'll try what you proposed 👍

girardinsamuel:
@josephmancuso I made required changes 👍

josephmancuso (approved):

josephmancuso:
Looks great! that worked perfectly

josephmancuso (approved):

PR #361, opened 2021-02-03
Fixed error when was used the enum type in postgres

### LINKED ISSUE #359: Enum in migration not working (Postgres)
**Describe the bug**
A migration using enum type in postgres database isn't working.

**To Reproduce**
1. Create a migration with `table.enum('gender', ['male', 'female'])`
2. Run migration

**Expected behavior**
Constraint check is have to create normally.

**Screenshots**
image

Postgres

Marlysson:
Closes #359 
@josephmancuso It's done and I had to change somethings.

1. Removed the CHECK from mapped types because when the process of "columnize" start, then it was placed in `length` position ( acting as length of column ) for some way, making it impossible change it ( I think that was a bug ).
2. Added `values` in Constraint model to represent the values that the column will compare against.
3. Created constraint `get_[constraint]_string` in platform.

No break change with this changes.

### DIFF
--- a/src/masoniteorm/schema/Blueprint.py
+++ b/src/masoniteorm/schema/Blueprint.py
@@ -398,8 +398,16 @@ class Blueprint:
         for option in options:
             new_options += "'{}',".format(option)
         new_options = new_options.rstrip(",")
+
         self._last_column = self.table.add_column(
-            column, "enum", length=new_options, nullable=nullable
+            column, "enum", length=255, nullable=nullable
+        )
+
+        self.table.add_constraint(
+            self._last_column.name,
+            "check",
+            columns=[self._last_column.name],
+            values=new_options,
         )
         return self
 
--- a/src/masoniteorm/schema/Constraint.py
+++ b/src/masoniteorm/schema/Constraint.py
@@ -1,5 +1,6 @@
 class Constraint:
-    def __init__(self, name, constraint_type, columns=None):
+    def __init__(self, name, constraint_type, columns=None, values=None):
         self.name = name
         self.constraint_type = constraint_type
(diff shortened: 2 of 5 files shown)


src/masoniteorm/schema/Blueprint.py
```diff
@@ -398,8 +398,16 @@ def enum(self, column, options=None, nullable=False):
         for option in options:
             new_options += "'{}',".format(option)
         new_options = new_options.rstrip(",")
+
         self._last_column = self.table.add_column(
-            column, "enum", length=new_options, nullable=nullable
+            column, "enum", length=255, nullable=nullable
+        )
+
+        self.table.add_constraint(
```
josephmancuso, line 406: My only issue with this PR is adding a dedicated constraint type here. Won't things like sqlite, mssql and mysql need to specially handle this? What happens when those databases don't have check constraints?

It's past midnight right now but I'm actually confused on why some of those tests didn't fail lol ...

josephmancuso (changes requested):

src/masoniteorm/schema/Blueprint.py
josephmancuso: I thought we were just going to add a dedicated column constraint. I did something like this in #363 to compare these 2 approaches.

Marlysson: @josephmancuso by that I saw, all databases ( that we support ) had the `check` constraint handled equally. This change  define in `table level` and don't `column level`. 

But I can change to what you do in SQLite.

josephmancuso:
Great try at this but will fix in #363

PR #363, opened 2021-02-03
fixed enum

josephmancuso:
Relates #359 

This fixes an issue with enum with SQLite.

@Marlysson after testing the postgres PR i realized there was an issue also with the sqlite enum. There is also no enum for sqlite but there is also a CHECK constraint just like Postgres


josephmancuso:
This also fixed postgres and mysql

josephmancuso:
Also fixed for mssql

PR #368, opened 2021-02-04
model: Implement get_dirty_keys()

resmo:


josephmancuso (dismissed):

girardinsamuel (dismissed):

girardinsamuel:
Looks like there is still an issue with the tests 🤔

tests/models/test_models.py
```diff
@@ -42,6 +42,7 @@ def test_model_original_and_dirty_attributes(self):
         self.assertEqual(model.get_original("username"), "joe")
         self.assertEqual(model.get_dirty("username"), "bob")
         self.assertEqual(model.__dirty_attributes__["username"], "bob")
+        self.assertEqual(list(model.get_dirty_keys()), ["username"])
```
josephmancuso: Should we make the keys return a list by default. I believe the way you have it right now it will return a dict_keys() structure. Which is why you're doing this weird thing in the test

josephmancuso (changes requested):

josephmancuso:
Tests are failing because models have a builder attribute. This needs to be popped off I think

josephmancuso:
See line 579 on the Model.py file

josephmancuso:
We prob need a dedicated method like get_dirty_attributes() which does the pop for us

src/masoniteorm/models/Model.py
```diff
@@ -570,6 +570,9 @@ def get_original(self, key):
     def get_dirty(self, key):
         return self.__dirty_attributes__.get(key)
 
+    def get_dirty_keys(self):
+        return list(self.get_dirty_attributes().keys())
```
josephmancuso, line 574: I guess we had this method already 😅

josephmancuso (approved):

josephmancuso (approved):

PR #375, opened 2021-02-06
add context manager for transaction

erhuabushuo:


josephmancuso:
This just needs a test around it. should be a test_transactions.py file somewhere

josephmancuso (approved):

PR #376, opened 2021-02-06
Add truncate to model 

### LINKED ISSUE #352: Add truncate() on models too
Right now we can do
```python
from config.database import DB
DB.get_schema_builder().truncate("licenses")
``` 
Would be cool to be able to do it on the model
```python
License.truncate()
```

girardinsamuel:
Fixes #352

I guess I am doing something wrong in the truncate method of the query builder


src/masoniteorm/query/QueryBuilder.py
```diff
@@ -1546,6 +1546,9 @@ def when(self, conditional, callback):
             callback(self)
         return self
 
+    def truncate(self, foreign_keys=False):
+        return self.get_schema_builder().truncate(self._table, foreign_keys=False)
```
girardinsamuel: maybe I should do instead ?
```python
self.grammar.truncate_table(self._table)
```

josephmancuso: Yeah exactly. We have to duplicate that sql statement in each query grammar class

josephmancuso: I don't think it will be a good idea to bind the query and schema classes together so we'll just have to duplicate sql in order to make it less rigid

src/masoniteorm/query/QueryBuilder.py
```diff
@@ -1546,6 +1546,12 @@ def when(self, conditional, callback):
             callback(self)
         return self
 
+    def truncate(self):
+        import pdb
+
+        pdb.set_trace()
```
josephmancuso: whats this?

tests/mysql/model/test_model.py
```diff
@@ -229,3 +229,6 @@ def test_serialize_with_dirty_attribute(self):
 
         def test_attribute_check_with_hasattr(self):
             self.assertFalse(hasattr(Profile(), "__password__"))
+
+        def test_can_truncate_table_from_model(self):
+            User.truncate()
```
josephmancuso: Shouldn't be testing the truncating, better to test the query builder query

josephmancuso (changes requested):
I just realized i never submitted this review

tests/mysql/model/test_model.py
girardinsamuel: truncate has already been tested I believe. Here I the idea was to check if we can indeed call truncate on the User model ? I did not complete the test but the idea was to check if a query was triggered when calling truncate on model.
What do you think ?

girardinsamuel: Right now `User.truncate()` raises 
`MySQLGrammar' object has no attribute 'truncate_table_string'`

girardinsamuel: Actually I have found out some missing methods ? 🤔 (or maybe I did not understood things)
in BaseGrammar I was expecting the following methods to be defined or defined in the different grammar class (SQLite, MySQL ...) but those methods are never defined:

```python
rename_table_string()
drop_table_string()
drop_table_if_exists_string()
truncate_table_string()
```

girardinsamuel: should we add for example on each grammar classes (if it's the same) this method ?
```python
def truncate_table_string(self):
    return "TRUNCATE {table}"
```

josephmancuso: yeah exactly that's what I was getting at before. We can prob add it to the base grammar if its all the same and if there's any differences we can override the method at the grammar level

girardinsamuel: okay sorry did not get it immediately (I was thinking this was already done 😅)

josephmancuso: Yeah it prob used to be before we separated the DML queries from the DDL queries

girardinsamuel: @Marlysson  @josephmancuso 
I have added tests but somehow that's not working yet. I don't really understand why the sql is not picked up. It does not work for any of the grammars so the issue is in the implementation in the QueryBuilder I guess

src/masoniteorm/query/QueryBuilder.py
```diff
@@ -1546,6 +1546,10 @@ def when(self, conditional, callback):
             callback(self)
         return self
 
+    def truncate(self):
+        self.get_grammar().truncate_table(self._table)
```
josephmancuso: You might have to do something like

```python
def truncate(self):
     sql = self.get_grammar().truncate_table(self._table)
     if self._dry:
        return sql

     self.new_connection().query(sql, ())
```

Marlysson: @girardinsamuel That's the solution. When the querybuilder are in "dry" mode it can return just the sql generated, instead it hit the database.

girardinsamuel: - I am a bit surprised that this kind of logic is not present for other type of queries (drop, rename, ...) ?
- Also does this assume that query builders tests are running with dry mode enabled ?

Marlysson: 1. In Schema.py, the methods: truncate, drop_if_exists and truncate there the verification of the dry mode or to hit database, just the `drop` don't.
2. Once the Schema or QueryBuilder is built, there a flag to indicate this in constructor in tests, I think.

josephmancuso: > In Schema.py, the methods: truncate, drop_if_exists and truncate there the verification of the dry mode or to hit database, just the drop don't.

that's correct here. The code is built to be tested first so there are flags we pass to make things dry so we don't need a database just to test queries. If the queries we assert are correct then the actual database calls will 100% work. So no need to make complexity by testing actual database connections when testing the query that will run is just as effective.

> Once the Schema or QueryBuilder is built, there a flag to indicate that in constructor of them in tests I think.

also correct. The tests pass dry=True in the test case classes query builder initializer. Most test case classes have a get_builder() method where we set the dry parameter to `True`. Some might not but that's because those ones are just building a query and asserting the SQL. 

There are also some methods like I think create and update that take a query=True parameter that does the same thing but this is largely an inconsistency in the codebase

girardinsamuel:
Tests are passing now but I guess that's worth a new review. Also I fixed truncate in sqlite schema as `TRUNCATE` does not exist in SQLite.

src/masoniteorm/query/grammars/MySQLGrammar.py
```diff
@@ -219,3 +219,6 @@ def where_null_string(self):
 
     def where_not_null_string(self):
         return " {keyword} {column} IS NOT NULL"
+
+    def truncate_table_string(self):
+        return "TRUNCATE TABLE {table}"
```
josephmancuso: I just realized we need to support the whole enable / disable foreign keys thing. Tables won't be able to be truncated if they have foreign key constraints to other tables.  Typically when you want to truncate a table you want to disable this check.

See the platform classes as an example. We can honestly just copy and paste the method from the schema platform classes

josephmancuso: See here:

```python
return [
            self.disable_foreign_key_constraints(),
            f"TRUNCATE {self.wrap_table(table)}",
            f"DELETE FROM {self.wrap_table(table)}",
            self.enable_foreign_key_constraints(),
        ]
```

josephmancuso (changes requested):

src/masoniteorm/query/grammars/MySQLGrammar.py
girardinsamuel: I added foreign keys handling for SQLite only (just to check), is this what you meant ?

I found weird that passing foreign_keys=True will disable foreign_keys handling I would have done it the other way around but that's how it's done in Schema. 🤔

josephmancuso:
@girardinsamuel tests are failing again?

src/masoniteorm/query/grammars/MySQLGrammar.py
girardinsamuel: @josephmancuso I posted this above. I updated only SQLite so that is why other tests are failing. Can you confirm what I have done is okay for SQLite ?

src/masoniteorm/query/grammars/SQLiteGrammar.py
```diff
@@ -179,3 +179,22 @@ def value_equal_string(self):
...
+
+    def truncate_table_string(self, foreign_keys=False):
+        # SQLite do not have TRUNCATE TABLE command but we can
+        # use SQLite DELETE command to delete complete data from an existing table
+        if not foreign_keys:
+            return "DELETE FROM {table}"
+
+        return (
+            f"{self.disable_foreign_key_constraints()};"
+            + "DELETE FROM {table};"
```
josephmancuso: Yeah this is exactly right

src/masoniteorm/query/grammars/MySQLGrammar.py
josephmancuso: Oh sorry I didn't see that. Yeah what you did in sqlite is correct

src/masoniteorm/query/grammars/MSSQLGrammar.py
```diff
@@ -166,3 +166,13 @@ def table_insert_column_string(self):
 
     def value_string(self):
         return "'{value}'{separator}"
+
+    def truncate_table_string(self, foreign_keys=False):
+        if not foreign_keys:
+            return "TRUNCATE TABLE {table}"
+
+        return (
+            "ALTER TABLE {table} NOCHECK CONSTRAINT ALL;"
```
josephmancuso: This should return a list instead of a semi colon separated string of sql statements

src/masoniteorm/query/grammars/MySQLGrammar.py
```diff
@@ -227,3 +227,18 @@ def where_null_string(self):
...
+
+    def disable_foreign_key_constraints(self):
+        return "SET FOREIGN_KEY_CHECKS=0"
+
+    def truncate_table_string(self, foreign_keys=False):
+        if not foreign_keys:
+            return "TRUNCATE TABLE {table}"
+
+        return (
+            f"{self.disable_foreign_key_constraints()};" + "TRUNCATE TABLE {table};"
```
josephmancuso: same thing here. this should look more like:

```python
return [
    self. disable_foreign_key_constraints(),
    "TRUNCATE TABLE {table}",
    self. enable_foreign_key_constraints()
]
```

src/masoniteorm/query/grammars/PostgresGrammar.py
```diff
@@ -184,3 +184,13 @@ def where_null_string(self):
...
     def where_not_null_string(self):
         return " {keyword} {column} IS NOT NULL"
+
+    def truncate_table_string(self, foreign_keys=False):
+        if not foreign_keys:
+            return "TRUNCATE TABLE {table}"
+
+        return (
+            "ALTER TABLE {table} DISABLE TRIGGER ALL;"
+            + "TRUNCATE TABLE {table};"
```
josephmancuso: same thing here

src/masoniteorm/query/grammars/SQLiteGrammar.py
```diff
@@ -179,3 +179,21 @@ def value_equal_string(self):
...
+        return "PRAGMA foreign_keys = OFF"
+
+    def truncate_table_string(self, foreign_keys=False):
+        # SQLite do not have TRUNCATE TABLE command but we can
+        # use SQLite DELETE command to delete complete data from an existing table
+        if not foreign_keys:
+            return "DELETE FROM {table}"
+
+        return (
+            f"{self.disable_foreign_key_constraints()};"
```
josephmancuso: same thing here

src/masoniteorm/schema/platforms/SQLitePlatform.py
```diff
@@ -319,11 +319,11 @@ def compile_column_exists(self, table, column):
 
     def compile_truncate(self, table, foreign_keys=False):
         if not foreign_keys:
-            return f"TRUNCATE {self.wrap_table(table)}"
+            return f"DELETE FROM {self.wrap_table(table)}"
 
         return [
             self.disable_foreign_key_constraints(),
-            f"TRUNCATE {self.wrap_table(table)}",
+            f"DELETE FROM {self.wrap_table(table)}",
```
josephmancuso, line 327: Like this

tests/mysql/builder/test_query_builder.py
```diff
@@ -25,14 +25,15 @@ def articles(self):
 class BaseTestQueryBuilder:
     maxDiff = None
 
-    def get_builder(self, table="users"):
+    def get_builder(self, table="users", dry=False):
```
josephmancuso: These should all be defaulting to dry=True

tests/mysql/builder/test_query_builder.py
```diff
@@ -842,3 +859,17 @@ def where_not_like(self):
...
+        builder.truncate()
+        """
+        return """TRUNCATE TABLE `users`"""
+
+    def truncate_without_foreign_keys(self):
+        """
+        builder = self.get_builder()
+        builder.truncate()
+        """
+        return """SET FOREIGN_KEY_CHECKS=0;TRUNCATE TABLE `users`;SET FOREIGN_KEY_CHECKS=1"""
```
josephmancuso: should be returning a list here

josephmancuso (changes requested):
Almost there! If i remember correctly, you can't pass a semi colon separated string of sql statements. Some of these connection classes don't support this. This is one of the reasons we support passing a list of SQL statements. If you look at any of the connection classes, we are checking if the query is an instance of a list

josephmancuso (approved):
@girardinsamuel fixed the issue with this PR here. Can review it to see how i did it. Not sure how the foreign key stuff i did yesterday got into this PR though 🤔

PR #389, opened 2021-02-16
Fix missing registered observer command

girardinsamuel:
`python craft observer ...` raises an error `observer command is not defined` because this ORM command is not registered


girardinsamuel:
@josephmancuso ready for review 👍 it's a really quick fix

josephmancuso (approved):

PR #393, opened 2021-02-16
Fill the model with internal attributes after create/saving

### LINKED ISSUE #390: Model no attribute 'id' after creating
**Describe the bug**

Storing a new model does not allow to access the `id` attribute. Accessing in code after `save()` results in a 
`class model 'Category' has no attribute id`

```python
category.save()
request.session.flash('success', f'Saved: {category.id}')
return request.redirect_to('admin.category.table')
```

**To Reproduce**

create a simple store function:

```python
from app.Category import Category

class CategoryController(Controller):

    def store(self, request: Request):
        category = Category()
        category.name = request.input('name')
        category.slug = request.input('slug')
        category.description = request.input('description')
        category.sorting = request.input('sorting')
        category.save()
        return category
```

**Expected behavior**
What do you believe should be happening?
```json
{
  "id": 1,
  "name": "test",
  "slug": "test",
  "description": "test",
  "sorting": "1"
}
```

**Actual behavior**
```json
{
  "name": "test",
  "slug": "test",
  "description": "test",
  "sorting": "1"
}
```

 - Type: SQLite 2.8.17
 - Masonite ORM:  masonite-orm==1.0.27

Marlysson:
Closes #390 
@josephmancuso As this is too simple, I submitted soon.

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -590,12 +590,14 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
             else:
                 result = self.create(self.__dirty_attributes__, query=query)
             self.observe_events(self, "saved")
+            self.fill(result.__attributes__)
             return result
 
         if self.is_loaded():
             result = builder.update(self.__dirty_attributes__, dry=query).to_sql()
         else:
             result = self.create(self.__dirty_attributes__, query=query)
+
         return result
 
     def get_value(self, attribute):


josephmancuso:
Did that fix the issue?

Marlysson:
yeah it does.

src/masoniteorm/models/Model.py
```diff
@@ -590,12 +590,14 @@ def save(self, query=False):
             else:
                 result = self.create(self.__dirty_attributes__, query=query)
             self.observe_events(self, "saved")
+            self.fill(result.__attributes__)
```
josephmancuso, line 593: I don't think it is `__attributes__` right? Because the result of the create will contain the `id` ... 🤔

josephmancuso (changes requested):

src/masoniteorm/models/Model.py
Marlysson, line 593: Passing the whole "result" doesn't work, it receive a dictionary.

josephmancuso: shouldn't it be:

```python
self.fill(result)
```

from the result of the create, which contains the `id` of the create we just created

josephmancuso (commented):
OOHH I see. you are doing `result.__attributes__` because it is a model that gets returned ..

src/masoniteorm/models/Model.py
Marlysson: When I used the self.fill(result) it raises this in `fill` method:

```python
self.__attributes__.update(attributes)
...
TypeError: getattr(): attribute name must be string
```

If I'm not wrong the type of "result" is a Model ( Category) , then I use the atributes returned by result.

Marlysson:
Exactly. There other cases?

josephmancuso:
No not when working with models I think this works perfect. Its been a long day lol. I was reading it as `self.__attributes__` instead of `result.__attributes__`. My eyes are tired

josephmancuso (approved):

PR #405, opened 2021-03-05
fix __casts__ bug

erhuabushuo:


src/masoniteorm/models/Model.py
```diff
@@ -631,14 +633,14 @@ def get_cast_map(self):
         cast_map.update(self.__cast_map__)
         return cast_map
 
-    def _cast_attribute(self, attribute):
```
erhuabushuo, line 634: here cast the attribute name, not the actual value.

josephmancuso:
thanks! Added some more tests

PR #423, opened 2021-03-24
Feature/422

### LINKED ISSUE #422: Allow other database objects to have their name explicitly specified (eg indexes, foreign keys)
Currently, Masonite ORM generates database object names by appending some hard coded values when generating the DDL from the migrations.  I need to follow a different pattern to comply with client-specified database naming norms.

For example,
`table.string("email").unique()` on the `users` table definition leads to a unique index with the database object name of `users_email_unique`

I would like to be able to specify a param on `unique()`, `index` and `foreign` which , if present, would be used instead of the auto generated name:
```
table.unique(column="email", name="uq_users_email")
table.index(column="some_other_field", name="ix_users_some_other_field")
table.foreign(column="user_type_id", name="fk_users_user_type_id").references("id").on("user_types")
```
etc

Thanks!

josephmancuso:
Closes #422 

This PR adds a high level way to change constraint names without needing to use low level methods.

For indexes:

```python
table.unique(["col1", "col2"], name="unique_idx")
table.index(["col1", "col2"], name="normal_idx")
table.foreign("col", name="col_foreign").references(..).on(..)
```

### DIFF
--- a/src/masoniteorm/schema/Blueprint.py
+++ b/src/masoniteorm/schema/Blueprint.py
@@ -628,7 +628,7 @@ class Blueprint:
     def soft_deletes(self, name="deleted_at"):
         return self.datetime(name, nullable=True).nullable()
 
-    def unique(self, column=None):
+    def unique(self, column=None, name=None):
         """Sets the last column to be unique if no column name is passed.
 
         If a column name is passed this method will create a new unique column.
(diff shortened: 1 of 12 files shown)


josephmancuso (commented):
reviewed

PR #425, opened 2021-03-24
added support for adding indexes without needing column name

### LINKED ISSUE #424: Make column name optional when specifying index/foreign/fulltext inline
This is a follow on to #422

If I use method chaining on the field as it is being created, the index() method expects the column param:

`table.integer(column="user_id").unsigned().index(name='ix_profiles_user_id')`

Gives:
```
  TypeError

  index() missing 1 required positional argument: 'column'
```

This may be confusing and although
`table.integer(column="user_id").unsigned().index(column='user_id', name='ix_profiles_user_id')` 
works, I think it is a little clunky to have to specify the column name twice in the same definition.

Would it be possible to make the `column` attribute optional for `index()`, `foreign()` and `fulltext()`?  `unique()` is already optional, so is fine.

When creating the index/FK on separate lines, the column name of course would still be required.

Many thanks

josephmancuso:
Closes #424

### DIFF
--- a/src/masoniteorm/schema/Blueprint.py
+++ b/src/masoniteorm/schema/Blueprint.py
@@ -636,11 +636,8 @@ class Blueprint:
         Returns:
             self
         """
-        if column is None:
-            self.table.add_constraint(
-                self._last_column.name, "unique", columns=[self._last_column.name]
-            )
-            return self
+        if not column:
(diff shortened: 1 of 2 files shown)


josephmancuso (commented):
reviewed

PR #426, opened 2021-03-26
Adding add_foreign function

### LINKED ISSUE #388: Add `add_foreign` to the schema builder
We should have an `add_foreign` method on the schema builder. This should be a shortcut to the current way to add foreign keys. It should look like this:

```python
table.add_foreign("from_column.to_column.table")
```

The above would be the same as this:

```python
table.foreign("from_column").references("to_column").on("table")
```

alfonsocv12:
Closes #388 

This is the simples way that I could think of doing this, let me know what i need to change

### DIFF
--- a/src/masoniteorm/schema/Blueprint.py
+++ b/src/masoniteorm/schema/Blueprint.py
@@ -716,12 +716,12 @@ class Blueprint:
         Arguments:
             columns {string} -- The name of the from_column . to_column . table
         """
-        columns_array: list = columns.split(".")
-        if len(columns_array) != 3:
+        if len(columns.split(".")) != 3:
             raise Exception("Wrong add_foreign argument, the struncture is from_column.to_column.table")
-        self.foreign(columns_list[0])
-        self.references(columns_list[1])
-        self.on(columns_list[2])
+        from_column, to_column, table = columns.split(".")
+        self.foreign(from_column)
+        self.references(to_column)
+        self.on(table)
         return self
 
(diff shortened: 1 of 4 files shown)


src/masoniteorm/schema/Blueprint.py
```diff
@@ -709,6 +709,21 @@ def primary(self, column=None):
...
         return self
 
+    def add_foreign(self, columns: string):
+        """Creates the foreign spliting the foreign name, reference column, and
+        reference table.
+
+        Arguments:
+            columns {string} -- The name of the from_column . to_column . table
+        """
+        columns_array: list = columns.split(".")
```
josephmancuso: I found using unpacking is more clear.

Changing this line to 
```python
        from_column, to_column, table = columns.split(".")
```

makes things a little more clear on what is going on

josephmancuso (changes requested):
Perfect! pretty simple ticket just a small change I would make to make the code clearer

josephmancuso:
Also should add a test in one of the grammar test classes like `tests/mysql/schema/test_mysql_schema_builder.py`

src/masoniteorm/schema/Blueprint.py
alfonsocv12: And remove the validation down on leve it ?

josephmancuso (approved):
Looks great! I'l manually test this when I get time later but looks good.

josephmancuso (approved):

PR #433, opened 2021-04-03
Add only changed attributes during updates and add force mode

girardinsamuel:
When updating a record with `update()` method, if some attributes are given but have not changed, they will be added in the SQL `UPDATE` query. This can lead to some issues.

This PR changes the behaviour to remove unchanged attributes during the update and allow to force the old behaviour in 3 ways:
- with a `force=True` argument
```python
user.update({"email": "test@test.fr", "username": "Sam"}, force=True)
```
- There is also a new `force_update()` method on class/query builder that will always force the update with all the attributes given in argument:
```python
user.force_update({"email": "test@test.fr", "username": "Sam"})
```
- Finally a new attribute `__force_update__` attribute is available on model class that always enable forcing to avoid specifying it each time in `update()` method. The default value is `False`.
```python
class User(Model):
     __force_update__ = True
```

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -106,6 +106,7 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
     __timestamps__ = True
     __timezone__ = "UTC"
     __with__ = ()
+    __force_update__ = False
 
     date_created_at = "created_at"
     date_updated_at = "updated_at"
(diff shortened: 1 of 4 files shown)


josephmancuso (commented):
I haven't yet tested this PR but keep in mind that the query that generates may be wrong if trying to update values where nothing changes. Like updating in that original example of an email. If the only thing you updated was the email but the value didn't change then you would be passing along an empty dictionary which I feel could break things. May not be the case but something to test for

girardinsamuel:
@josephmancuso @Marlysson it's ready for review. I normally added tests to cover the different use cases.

But for now this new feature is applied only when using models and not when using only query builder... is it ok ?

josephmancuso (approved):
Works perfectly!

PR #435, opened 2021-04-04
Adding seeder command after migrate:refresh command.

### LINKED ISSUE #434: Ability to run seeders in refresh migrations
Some way to run seeders in same command of migrate:refresh.

Run all seeders
```
masonite-orm migrate:refresh --seed
```

Run specific seeder
```
masonite-orm migrate:refresh --seed CustomSeeder
```

Related tasks:
- [ ] Update documentation

Call seeder command inside refresh migration command

Marlysson:
Close #434


Marlysson:
@girardinsamuel All done.
Can you review? After that I'll add this option in documentation.

src/masoniteorm/commands/SeedRunCommand.py
```diff
@@ -15,13 +15,26 @@ class SeedRunCommand(Command):
...
         if self.argument("table") == "None":
-            return Seeder(
-                dry=self.option("dry"), seed_path=self.option("directory")
-            ).run_database_seed()
+            seeder.run_database_seed()
+            seeder_seeded = "Database Seeder"
+
+        else:
+
+            table = self.argument("table")
```
girardinsamuel, line 27: I find it simplier to just do (even if there is one repetition)
```suggestion
            table = self.argument("table")
            seeder_file = f"{underscore(table)}_table_seeder.{camelize(table)}TableSeeder"
            seeder.run_specific_seed(seeder_file)
            seeder_seeded = f"{camelize(table)}TableSeeder"
```

src/masoniteorm/commands/MigrateRefreshCommand.py
```diff
@@ -9,12 +9,22 @@ class MigrateRefreshCommand(Command):
     migrate:refresh
         {--c|connection=default : The connection you want to run migrations on}
         {--d|directory=databases/migrations : The location of the migration directory}
+        {--s|seed=? : Wheter seeding should be run after refresh migrations and which seedeer should be ran}
```
girardinsamuel: I would update the description to:
```suggestion
        {--s|seed=? : Seed database after refresh. The seeder to be ran can be provided in argument}
```

girardinsamuel (changes requested):
That's a nice little feature 👍

girardinsamuel (dismissed):

src/masoniteorm/commands/MigrateRefreshCommand.py
```diff
@@ -9,12 +9,22 @@ class MigrateRefreshCommand(Command):
     migrate:refresh
         {--c|connection=default : The connection you want to run migrations on}
         {--d|directory=databases/migrations : The location of the migration directory}
+        {--s|seed=? : Seed database after refresh. The seeder to be ran can be provided in argument}
```
josephmancuso, line 12: What does this `?` do?

src/masoniteorm/commands/MigrateRefreshCommand.py
```diff
@@ -9,12 +9,22 @@ class MigrateRefreshCommand(Command):
...
             command_class=self,
             connection=self.option("connection"),
             migration_directory=self.option("directory"),
         )
+
         migration.refresh()
+
+        self.line("")
+
+        if self.option("seed") == "null":
```
josephmancuso, line 27: When is this option set to `"null"`?

josephmancuso (changes requested):

src/masoniteorm/commands/MigrateRefreshCommand.py
Marlysson: Means an optional option

https://cleo.readthedocs.io/en/latest/introduction.html#using-options

Using just a option isn't possible If the absent If It it's because isn't define or define with no value

src/masoniteorm/commands/MigrateRefreshCommand.py
Marlysson: When option isn't defined it's comes "null", when it's defined but with no value then It comes None, otherwise it's  the value defined for option.

src/masoniteorm/commands/MigrateRefreshCommand.py
girardinsamuel: ah that's nice I always wondered how to have an optional arg without value but which can take value if needed 💪

josephmancuso (approved):

Issue #440, opened 2021-04-11
LONGBLOB datatype is not supported on PostgreSQL


rodrigopv:
**Describe the bug**
I'm re-creating an existing PostgreSQL database for my open-source project. I need to store small files, for which postgres uses the `bytea` data type. When looking into docs, the closes one we have is Binary, but it seems to try using a non-existent datatype for postgres when looking into the SQL output of `masonite-orm migrate -s`.

**To Reproduce**
1. Create migration with a binary filetype when using PostgreSQL database.
2. Execute or preview migration SQL to check the output
3. Expect to see PostgreSQL error `2021-04-11 21:49:07.609 UTC [530] ERROR:  type "longblob" does not exist at character 109` during migration

**Expected behavior**
ORM should try to create the column with `bytea` data type.

**Screenshots or code snippets**
Screenshots help a lot. If applicable, add screenshots to help explain your problem.
![image](https://user-images.githubusercontent.com/198869/114322631-475bec80-9aef-11eb-85d8-92ecf19ac26d.png)

**What database are you using?**
 - Type: Postgres + TimescaleDB (docker image:  timescale/timescaledb:2.1.1-pg13)
 - Version: 13
 - Masonite ORM `masonite-orm==1.0.45`

PR #443, opened 2021-04-11
feat/442: implement postgres specific inet, cidr and macaddr datatypes

### LINKED ISSUE #442: PostgreSQL inet,cidr and macaddr data type
Would be nice to be able to store IP addresses in inet data type

There's no way to manipulate IP addresses or IP networks in Masonite-ORM.

**Additional context**
ActiveRecord from Rails has 2 approaches when using `inet` and other networking-related data types:

1. map to existing ruby objects when available (like Ruby's IPAddr)
2. map to string when not

To implement this, 2 tasks can be considered:
- [x] Support using `inet`, `cidr` and `macaddr` datatype on migrations so we keep using Masonite on projects where this data type is needed. Raw queries might still be used for making use of the datatype related operators.
- [ ] Implement operators for `inet` and `cidr` datatype: this way we can eventually get rid of raw queries/prepared statements.
 
More information on datatypes: https://www.postgresql.org/docs/9.1/datatype-net-types.html

I'm pretty much new to this framework as I'm doing all kind of efforts to get rid of Rails, maybe a previous blocker would be to support database-exclusive data types, not sure how Masonite-ORM is implementing `jsonb` at the moment.

rodrigopv:
Closes #442 

This PR should enable using those datatypes on migrations so we can at least create tables with those wonderful datatypes. I was in doubt if I should proceed with this PR, but then I saw the "fulltext" implementation that is only for MySQL and this should be somewhat similar.

Adding an exclusive datatype should introduce a bug just like #440, but as I commented in #442, treating these values as string or VARCHAR on other DBMS should avoid the bug.

For further query builder implementation a more complex method of supporting these datatypes should be considered for DBMS other than PGSQL.

This PR would allow me to keep using migrations with this ORM and then do prepared statements or raw queries when I need to use related operators for these datatypes.


josephmancuso:
Looks great 👍 . definitely the way to implement it.

josephmancuso:
I don't think we need all the extra tests for the other grammars as there are plenty of tests for compiling those columns into varchars. Once we get the postgres tests working it will be good enough

rodrigopv:
Actually I found an interesting detail during the last tests development. Other DBMS VARCHAR replacements would require a default varchar length, so I'm implementing that part as well

josephmancuso (approved):
Looks good to me 👍

Issue #419, opened 2021-03-15
How to use **JSON** fields with **masonite-orm** ?


marcelovieirahonorio:
**Describe the bug**
Please, how to use a JSON fields with masonite-orm?
In __orator__, I had use a __dict__ to assign a value to JSON field but, this not work in masonite-orm and don't find in documentation what is the correct way.

**To Reproduce**
1. Create a JSON field in database with migration (PostgreSQL)
2. Create a Model and set the __cast__ to field (in orator the cast was _dict_ , in masonite-orm, the doc tell to make a cast to _json_
3. Create a dict variable
4. Set value to a dict
5. Set the dict var to model field
6. Save the model

**Expected behavior**
I think that the model could save the _dict_ value to database

**Screenshots or code snippets**
In image 01 , the sample working with masonite 2.3...
In image 02, the sample doesn't working with masonite 3.0 and cast to dict
In image 03, the sample doesn't working with masonite 3.0 and cast to json

![image](https://user-images.githubusercontent.com/56178491/111160241-ac680500-8578-11eb-9c52-c8fe6689db4b.png)

![image](https://user-images.githubusercontent.com/56178491/111160278-b4c04000-8578-11eb-9e18-5b0df83b5385.png)

![image](https://user-images.githubusercontent.com/56178491/111160310-b8ec5d80-8578-11eb-82c8-cf6a5440bb8c.png)

**Desktop (please complete the following information):**
 - OS: Linux
 - Version Ubuntu 20.04

**What database are you using?**
 - Type: PostgreSQL
 - Version 12.6
 - Masonite ORM 1.0.40

josephmancuso:
Found the issue. Issue was that the model was not casting the attribute before sending it to the connection class. 

Will push a release in a bit

marcelovieirahonorio:
Thank you, I will test it when published in the pypi :)

josephmancuso:
@marcelovieirahonorio this is done and on version 1.0.42

marcelovieirahonorio:
@josephmancuso , I tried from different forms, but, doesn't work :( :

with __cast__ to JSON:
![image](https://user-images.githubusercontent.com/56178491/114445463-7da87300-9ba6-11eb-8f1f-bcb207053037.png)


with __cast__ to DICT:
![image](https://user-images.githubusercontent.com/56178491/114445601-a6306d00-9ba6-11eb-87e2-3f639c68c3aa.png)


without __cast__:
[...]

PR #444, opened 2021-04-12
Cast from dict object to json field

marcelovieirahonorio:
Source of issue: https://github.com/MasoniteFramework/orm/issues/419   
#419 

When obtaining the object, the json field was already loaded as dict, however, JsonCast was forcing cast again.
To resolve this, a check has been added if the value is already dict. If true, return the value, otherwise, cast.

*first issue -> this is my first Pull Request 

image


josephmancuso (approved):

josephmancuso:
Look great!

PR #449, opened 2021-04-25
Fix/447

### LINKED ISSUE #447: For Postgres: 'QueryBuilder' object has no attribute 'where_between'
For Postgres: 'QueryBuilder' object has no attribute 'where_between':

`    def get_pg_data_date(self, tabelle, startDate, endDate):
       
        builder = QueryBuilder().table(tabelle)
        results = builder.table(tabelle).where_between('timestamp', startDate, endDate).get()
        return results
`

Marlysson:
Closes #447 
@josephmancuso @girardinsamuel This PR allow the use of the between and not_between methods in query builder and model level.

Currently occours the error of different bindings quantities when query between and not_between is persisted, this happen because instead the sql keeping with qmarks('?') to after be filled with values, the values is fill direct in query by its value:

Instead:
`select * from "users" where "age" between '?' and '?' bindings = [1,10]
`

Was:
`select * from "users" where "age" between '1' and '10' bindings = [1,10]
`

In this case, then the dbapi would define de bindings in the second case the code would broke. It waiting by 2 qmarks to fill with bindings but not one is found.

Then the solution was define the bindings just when isn't a "BetweenExpression" ( I think this is not so good, because of the values directly in query, and to change all code to uses the qmark so many things will be changing in code, we can do that step-by-step )


josephmancuso:
Needs the tests to be fixed

Marlysson:
@josephmancuso Done. Basically added:

```
builder.between()
builder.not_between()
Model.between()
Model.not_between()
Model.where_not_null()
```

josephmancuso (dismissed):

girardinsamuel (approved):

josephmancuso (approved):

PR #458, opened 2021-05-14
fix bind value bug

### LINKED ISSUE #456: Looks like binding a value of 0 does not add the query binding correctly
**Describe the bug**
Looks like binding a value of 0 does not add the query binding correctly:

```python
builder.where('column', 0)
```

does not seem to add the binding to the query

erhuabushuo:
Closes #456 

bind value may be 0， False, None


josephmancuso:
I fixed this in another PR but i actually like this fix a lot better. Great job!!

PR #466, opened 2021-06-10
Feature/465

### LINKED ISSUE #465: Date column causes pendulum exception
As mentioned on Slack earlier today, I have a Person model that is defined with a date column (not datetime). When trying to retrieve it with a query like ```users = User.with_("person", "name", "roles").order_by('last_active_at', 'desc').get()``` it throws an exception in Pendulum "instance() only accepts datetime objects." (I suspect it would do this even when querying the Person directly, but I haven't tried.)

The model has the date of birth (DOB) column indicated as a date type:

```python
class Person(Model):                                                             
    """Person Model."""                                                          
    __table__ = 'people'                                                         
    __dates__ = ['dob']
```   

For columns that are defined as date or datetime in the database, I haven't found it necessary to specify `__dates__` in the model, but I tried it here anyway. Perhaps I'm misunderstanding what this does or when it is needed. I suspect it is mainly for fields that are strings/char types holding date information.

 - Type: MySQL
 - Masonite ORM v1.0.48

**Additional context**
It looks like date types need to be handled in the get_new_date method before this line
https://github.com/MasoniteFramework/orm/blob/0354e62c844da9d57c2302acc9b611270bd26759/src/masoniteorm/models/Model.py#L734
 
Joseph mentioned the [pendulum issue #250](https://github.com/sdispater/pendulum/issues/250) which seems to be the problem. This could be fixed in either the Masonite ORM Model or in pendulum. 

Querying and using this date column was working fine with Masonite 2 using Orator ORM. (I am working on migrating the code to Masonite 3.) I believe Orator handled the date types before it did its "hydration" step. Perhaps it would be worthwhile to look at what it did.

josephmancuso:
Closes #465 
* Changes timestamp columns to default to NULL instead of defaulting to current timestamp
* Adds ability to add a primary key with an altering query
* Fixes issue where datetime objects could not be converted to pendulum objects

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -1,5 +1,5 @@
 import json
-from datetime import datetime
+from datetime import datetime, date as datetimedate, time as datetimetime
 
 from inflection import tableize
 import inspect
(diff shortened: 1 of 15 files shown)


josephmancuso (commented):
reviewed

PR #469, opened 2021-06-16
fixed pivot serializing

josephmancuso:
Need to work through a little bit more of this ..

The first thing is we need to hide relationship data somehow. For example, apart of the pivot information you have access to the pivot data 

Screen Shot 2021-06-15 at 11 18 56 PM

I feel you may or may not want to return this data. For example, to an API response.

To solve this i added a new property:

```python
__relationship_hidden__ = {"people": ['pivot']}
```

This hides the `pivot` key from the `people` pivot table serialization

This PR also fixes a few issues i found with the belongs to many relationship

### DIFF
--- a/src/masoniteorm/collection/Collection.py
+++ b/src/masoniteorm/collection/Collection.py
@@ -303,13 +303,13 @@ class Collection:
     def reverse(self):
         self._items = self[::-1]
 
-    def serialize(self):
+    def serialize(self, *args, **kwargs):
         def _serialize(item):
             if self.__appends__:
                 item.set_appends(self.__appends__)
 
             if hasattr(item, "serialize"):
-                return item.serialize()
+                return item.serialize(*args, **kwargs)
             elif hasattr(item, "to_dict"):
                 return item.to_dict()
             return item
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -449,6 +449,7 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
         """
         serialized_dictionary = self.__attributes__
 
+
         # prevent using both hidden and visible at the same time
         if self.__visible__ and self.__hidden__:
             raise AttributeError(
@@ -484,10 +485,13 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
 
         # Serialize relationships as well
         serialized_dictionary.update(self.relations_to_dict())
+
         for append in self.__appends__:
             serialized_dictionary.update({append: getattr(self, append)})
 
         for key, value in serialized_dictionary.items():
+            if hasattr(value, 'serialize'):
+                value = value.serialize()
             if isinstance(value, datetime):
                 value = self.get_new_serialized_date(value)
             if key in self.__casts__:
@@ -531,6 +535,7 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
                 if value is None:
                     new_dic.update({key: {}})
                     continue
+                    
                 new_dic.update({key: value.serialize()})
 
         return new_dic
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -1,6 +1,5 @@
 import json
 from datetime import datetime, date as datetimedate, time as datetimetime
-from os import remove
 
 from inflection import tableize
 import inspect
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -717,7 +717,7 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
         if key in self.__attributes__:
             del self.__attributes__[key]
             return True
-        
+
         return False
 
     def get_dirty_attributes(self):
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -713,6 +713,13 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
 
         return attributes
 
+    def delete_attribute(self, key):
+        if key in self.__attributes__:
+            del self.__attributes__[key]
+            return True
+        
+        return False
+
     def get_dirty_attributes(self):
         if "builder" in self.__dirty_attributes__:
             self.__dirty_attributes__.pop("builder")
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -1,5 +1,6 @@
 import json
 from datetime import datetime, date as datetimedate, time as datetimetime
+from os import remove
 
 from inflection import tableize
 import inspect
@@ -105,6 +106,7 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
     __casts__ = {}
     __dates__ = []
     __hidden__ = []
+    __relationship_hidden__ = {}
     __visible__ = []
     __timestamps__ = True
     __timezone__ = "UTC"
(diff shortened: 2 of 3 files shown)


yubarajshrestha:
> Need to work through a little bit more of this ..
> 
> The first thing is we need to hide relationship data somehow. For example, apart of the pivot information you have access to the pivot data
> 
> Screen Shot 2021-06-15 at 11 18 56 PM
> 
> I feel you may or may not want to return this data. For example, to an API response.
> 
> To solve this i added a new property:
> 
> ```python
> __relationship_hidden__ = {"people": ['pivot']}
> ```
> 
> This hides the `pivot` key from the `people` pivot table serialization
> 
> This PR also fixes a few issues i found with the belongs to many relationship

Hiding pivot is a good idea, most of the times we just don't need those fields. But in some cases we also add more extra fields in pivot table I will give you an example hopefully that will make sense.

```python
class User(Model):
    """User Model."""
    @belongs_to_many('user_id', 'department_id', 'id', 'id')
    def departments(self):
        from app.models.Department import Department
        return Department
```

```python
class Department(Model):
    """User Model."""
    @belongs_to_many('department_id', 'user_id', 'id', 'id')
    def users(self):
        from app.User import User
        return User
```

Following is the pivot table and simple we don't usually need any model for this separately because we just use ORM relationship features to get the pivot data from the related tables.

```python
with self.schema.create("department_user") as table:
    table.increments("id")
    table.unsigned_integer('department_id').index('department_id')
    table.foreign('department_id').references('id').on('departments').on_delete('cascade')
    table.unsigned_integer('user_id').index('user_id')
    table.foreign('user_id').references('id').on('users').on_delete('cascade')
    table.boolean('is_head').default(False)
    table.timestamps()
```

So in above pivot table there is one extra field **is_head**. The concept is, user can have involved in different departments and he/she can also be the the **department head** in multiple departments in this scenario **is_head** field in pivot table makes sense.
This is just a simple example but in ecommerce sites we encounter so many cases like this.

```python
User.with_('departments').get()
```

Expected result with above query:

```json
[{
    "id": 1,
    "departments": [
        {
            "id": 1,
            "pivot": {
                "user_id": 1,
                "department_id": 1,
                "is_head": true
            }
        },
        {
            "id": 2,
            "pivot": {
                "user_id": 1,
                "department_id": 2,
                "is_head": false
            }
        },
    ]
}]
```
Expected result after hiding pivot field:

```json
[{
    "id": 1,
    "departments": [
        {
            "id": 1,
        },
        {
            "id": 2,
        },
    ]
}]
```

I hope all the above examples will help to understand the scenario.

yubarajshrestha (approved):
How about adding extra pivot field?

yubarajshrestha (approved):

yubarajshrestha (dismissed):

josephmancuso:
The last part of this is getting the extra field information in

josephmancuso (commented):
reviewed

josephmancuso:
@yubarajshrestha 

The solution i came up with is to add a new `with_fields` attribute to the `belongs_to_many` relationship:

```python
    @belongs_to_many('user_id', 'department_id', 'id', 'id', with_fields=['is_head'])
    def departments(self):
        from app.models.Department import Department
        return Department
```

You can then loop through the relation and access the pivot data:

```python
for department in User.find(1).departments:
    department.pivot.is_head #== 1 or 0
```

yubarajshrestha:
> @yubarajshrestha
> 
> The solution i came up with is to add a new `with_fields` attribute to the `belongs_to_many` relationship:
> 
> ```python
>     @belongs_to_many('user_id', 'department_id', 'id', 'id', with_fields=['is_head'])
>     def departments(self):
>         from app.models.Department import Department
>         return Department
> ```
> 
> You can then loop through the relation and access the pivot data:
> 
> ```python
> for department in User.find(1).departments:
>     department.pivot.is_head #== 1 or 0
> ```

Solution looks nice, I will give it a try, thank you.

yubarajshrestha:
@josephmancuso looks like pivot serialization issue has been fixed. But there are still few bugs,

```json
{
  "id": 4,
  "company_id": null,
  "name": "Random User",
  "code": "USR00003",
  "email": "abc@xyz.com",
  "password": "$2b$12$Ww.AQyCRVTMPEUo65iFSEObUxM3EGQyeqkSrCd4CZR5q.dTTyzTVe",
  "remember_token": null,
  "verified_at": null,
  "status": "active",
  "status_remarks": null,
  "created_at": "2021-06-20T11:08:26+00:00",
  "updated_at": "2021-06-20T11:08:26+00:00",
  "role_id": 2,
  "role": {
    "id": 2,
    "company_id": null,
    "name": "Admin",
    "slug": "admin",
    "description": "",
    "status": "active",
    "created_at": "2021-06-20T10:54:01.855877+00:00",
    "updated_at": "2021-06-20T10:54:01.855877+00:00"
  },
  "departments": [{
      "id": 1,
      "company_id": null,
      "name": "Engineering",
      "status": "active",
      "created_at": "2021-06-20T11:07:37+00:00",
      "updated_at": "2021-06-20T11:07:37+00:00",
      "user_id": 4,
      "department_id": 2,
      "pivot": {
        "user_id": 4,
        "department_id": 2,
        "id": 1,
        "is_head": true
      }
  },
  {
    "id": 2,
    "company_id": null,
    "name": "Accounting",
    "status": "active",
    "created_at": "2021-06-20T11:07:27+00:00",
    "updated_at": "2021-06-20T11:07:27+00:00",
    "user_id": 4,
    "department_id": 1
  }]
}
```

as you can see on the above sample pivot is missing in the second object. And also the pivot object sometimes it shows there and sometimes doesn't.

josephmancuso:
That second record definitely has a pivot record?

yubarajshrestha:
> That second record definitely has a pivot record?

Yes, as you can see, user belongs to multiple departments. if user id is 1, and the two department ids are 5 and 6 respectively then in database pivot it is stored as [user_id, department_id] i.e, [1,5] and [1,6] right?

josephmancuso:
Or should that pivot record be an empty dictionary?

yubarajshrestha:
> Or should that pivot record be an empty dictionary?

Data are stored as I explained in the above message. so empty dictionary? I didn't get it.

josephmancuso:
@yubarajshrestha I can't replicate this.

When i said should it be an empty dictionary i was saying does the 2nd user have a record inside the pivot table.

If they do not have a pivot record that is the only way i can replicate the issue.

Anyway you can provide more info? Like database screenshots of the users table, pivot table and foreign table?

yubarajshrestha:
> @yubarajshrestha I can't replicate this.
> 
> When i said should it be an empty dictionary i was saying does the 2nd user have a record inside the pivot table.
> 
> If they do not have a pivot record that is the only way i can replicate the issue.
> 
> Anyway you can provide more info? Like database screenshots of the users table, pivot table and foreign table?

I think you misunderstood. There's only one user in database and has relations with departments.

josephmancuso:
Maybe. Let's step back a bit.

> as you can see on the above sample pivot is missing in the second object. 

But you just said the user does not have a pivot record. The second user should not have a pivot record as they don't have a department ..

yubarajshrestha:
> Maybe. Let's step back a bit.
> 
> > as you can see on the above sample pivot is missing in the second object.
> 
> But you just said the user does not have a pivot record. The second user should not have a pivot record as they don't have a department ..

:) Let me explain it to you properly.

1. There's only one user in users table.
2. There are multiple records in departments table.
3. The only user we had in database is involved in 2 departments. 
4. Pivot table has two records. 
5. In above json output, you can see a single user has 2 departments, on which first department has pivot shown but in second one it is not shown.
6. Above json is output of only one user with role and departments. for which the query was: 

```python
result = User.with_("role", "departments").first()
```

josephmancuso:
gotchaaa. I See now lol. Thanks. I'll take a look then

josephmancuso:
@yubarajshrestha Thanks for the detailed feedback. The issue was that somehow i hard coded `pivot` as an attribute that should not be serialized .. must have forgot to remove it during testing.

This is fixed in version 1.0.52

yogesh-acharya:
Was this fix removed in latest upgrades ,, having with_fields in belongs to many relation does not give those fields .

PR #474, opened 2021-07-05
fixed issue for mysql not saving unique constraint

### LINKED ISSUE #472: unique() method in migrations is ignoring the name attribute
**Describe the bug**
Hi Joseph, the recent change to the ORM to allow indexes to be named does not seem to work for `unique()` - the name I specify is always ignored. `index()` on the other hand works fine. I tried diving down into the code down to the `Table.py` class, but I could not see anything obviously wrong.

**To Reproduce**
For example, as a side project I have started a rewrite of an old (2005) online game of mine.  Here is a simplified version (stripped to just the pertinent fields) of the migration for the games table:

```python
...
class CreateGamesTable(Migration):
    def up(self):
        """
        Run the migrations.
        """
        with self.schema.create("games") as table:
            table.increments("id")
            table.string("game_title", length=255, nullable=False)
            table.timestamps()
            # table.index(column="game_title", name="ix_games_game_title")  # <= this works when uncommented
            table.unique(column="game_title", name="uq_games_game_title")   # <= this does not
...
```

**Expected behavior**
What do you believe should be happening?
The field `games.game_title` should have a unique index on it called `uq_games_games_title`

However, although it does have a unique index, it's ignoring the supplied name and using the default:

```sql
SELECT DISTINCT
    TABLE_NAME,
    COLUMN_NAME,
    INDEX_NAME
FROM INFORMATION_SCHEMA.STATISTICS
WHERE TABLE_SCHEMA = 'new_dev'
  AND TABLE_NAME='games';

TABLE_NAME  COLUMN_NAME  INDEX_NAME
games       game_title   games_game_title_unique
games       id           PRIMARY
```

 - OS: macOS
 - Version 10.14.6

 - Type: MySQL
 - Version 8.0.21
 - Masonite ORM 1.0.52
 - PyMySQL 1.0.2

- Python 3.9.6
 
**Additional context**
N/A

josephmancuso:
Closes #472 

@jrouffiac Issue here was just that the mysql platform was building own own index name. This was the old way before the addition of specifying your own index names. Must have just missed the unique on the refactor.

This was also an issue for the other platforms as well (postgres and MSSQL)

### DIFF
--- a/src/masoniteorm/schema/platforms/MSSQLPlatform.py
+++ b/src/masoniteorm/schema/platforms/MSSQLPlatform.py
@@ -284,7 +284,7 @@ class MSSQLPlatform(Platform):
         return "CONSTRAINT {constraint_name} PRIMARY KEY ({columns})"
 
     def get_unique_constraint_string(self):
-        return "CONSTRAINT {table}_{name_columns}_unique UNIQUE ({columns})"
+        return "CONSTRAINT {constraint_name} UNIQUE ({columns})"
 
     def compile_table_exists(self, table, database):
         return f"SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = '{table}'"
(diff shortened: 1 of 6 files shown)


josephmancuso (commented):
reviewed

PR #486, opened 2021-08-06
Fix for scopes/UUIDPrimaryKeyScope.py always replacing primary_key

garzola:
Fix for scopes/UUIDPrimaryKeyScope.py always replacing the primary_key.  It should only provide a primary_key when one has not been provided.

### DIFF
--- a/src/masoniteorm/scopes/UUIDPrimaryKeyScope.py
+++ b/src/masoniteorm/scopes/UUIDPrimaryKeyScope.py
@@ -23,7 +23,9 @@ class UUIDPrimaryKeyScope(BaseScope):
         return str(uuid_func(*args))
 
     def set_uuid_create(self, builder):
-        uuid_version = getattr(builder._model, "__uuid_version__", 4)
-        builder._creates.update(
-            {builder._model.__primary_key__: self.generate_uuid(builder, uuid_version)}
-        )
+        # if there is already a primary key, no need to set a new one
+        if not builder._model.__primary_key__:
+            uuid_version = getattr(builder._model, "__uuid_version__", 4)
+            builder._creates.update(
+                {builder._model.__primary_key__: self.generate_uuid(builder, uuid_version)}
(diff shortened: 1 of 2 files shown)


garzola:
This is for Issue #486

src/masoniteorm/scopes/UUIDPrimaryKeyScope.py
```diff
@@ -23,7 +23,9 @@ def generate_uuid(self, builder, uuid_version):
         return str(uuid_func(*args))
 
     def set_uuid_create(self, builder):
-        uuid_version = getattr(builder._model, "__uuid_version__", 4)
-        builder._creates.update(
-            {builder._model.__primary_key__: self.generate_uuid(builder, uuid_version)}
-        )
+        # if there is already a primary key, no need to set a new one
+        if not builder._model.__primary_key__:
```
josephmancuso: This conditional is checking the model has a primary key. That is wrong. This should be replaced with:

```suggestion
        if builder._model.__primary_key__ not in builder._creates:
```

josephmancuso (changes requested):

src/masoniteorm/scopes/UUIDPrimaryKeyScope.py
josephmancuso: Thats why the tests are failing

garzola:
After looking carefully at this, I have a much better understanding of what
was happening.  Thanks for the comment.  I have updated the pull request
with this change.

I appreciate you taking the time to explain what was really needed.

Thanks,

Gustavo

--
"We must be willing to let go of the life we've planned, so as to have the
 life that is waiting for us."
                                   - Joseph Campbell

josephmancuso:
Pushed a small test fix to this branch to pass a test

josephmancuso (approved):
Looks good to me!

PR #495, opened 2021-08-16
fix join on value

circulon:
This PR properly addresses using arbitrary values with joins.

The where clause has been aliased (with a deprecation message) to on_value as where is not a valid sub-clause of Join and may build invalid syntax under certain circumstances.

This PR also fixes an issue where consecutive NULL/NOT NULL clauses would not be separated correctly 

"AND" methods added:
on_value()
on_is_null()
on_is_not_null()

"OR" methods added:
or_on_value()
or_on_is_null()
or_on_is_not_null()


josephmancuso:
@circulon just so we're on the same page what do you mean by?

> The where clause were removed as it cannot be used within the scope of a Join.

At least on MySQL those test cases are right with the `where` clause. They also actually work with the `and` clause you changed it too as well but what do you mean a `where` can't be used within the scope of a join?

circulon:
Hi @josephmancuso 

Yes you are correct in that using a where with a join is valid. I am not disputing that.
My point is the  where is not part of the filtering scope of a join it is part of the larger querys filtering scope.

This stackoverflow post explains it better than I can.
https://stackoverflow.com/questions/354070/sql-join-where-clause-vs-on-clause

The mysql 8.0 join syntax also notes
https://dev.mysql.com/doc/refman/8.0/en/join.html
```
join_specification: {
    ON search_condition
  | USING (join_column_list)
}
```
I hope this clarifies things for you?

Cheers

josephmancuso:
According to that stack overflow question, which I read quickly, it looks like you can have both depending on the join clause? 

What I'm really trying to get at is I don't think we should remove the `where` methods I think we should just add your new `on _` methods.

josephmancuso:
and even if we do remove those methods thats actually a breaking change

circulon:
Hi @josephmancuso 

Both Where and Join are clauses that are part of a select statement.
Where is not a part of a join clause.

The stack overflow post was showing the differences when using a select between where filtering and join ON filtering, so matbe that was the wrong context.

Besides the fact that using the where in a join like this creates invalid syntax.
In the current version of masonite orm you can do something like this
```python
builder = (
    QueryBuilder()
    .select("a, table_x.b , table_q.c")
    .from_("table_a")
    .join("table_x", lambda join: (
        (
            join.on("table_a_id", "=", "id")
           .where("dept", "=", "hr")
        )
    )
    .join("table_q", lambda join: (
        (
            join.on("table_a_id", "=", "id")
           .where("dept", "=", "accounts")
        )
    )
    .where("id", "=", "123456")
)
```
This will produce  an invalid query that looks similar to this
```sql
SELECT a, table_x.b, table_q.c
FROM table_a
JOIN table_x 
    ON id=table_a_id
    WHERE dept="hr"
JOIN table_q 
    ON id=table_a_id
    WHERE dept="accounts"
 WHERE id="123456"
```
This is invalid as there should only be one WHERE per SELECT.
Are you maybe thinking of the join as a sub-select and hence can use where in this fashion?

If you did want the where filtering across the main select as is implied by the above example it could be written in a much clearer form similar to this, which in this example is also laid out in the same way that you would actually write a raw query. 
```python
builder = (
    QueryBuilder()
    .select("a, table_x.b , table_q.c")
    .from_("table_a")
    .join("table_x", "table_a_id", "=", "id")
    .join("table_q", "table_a_id", "=", "id")
    .where("id", "=", "123456")
    .where("dept", "=", "hr")
    .or_where("dept", "=", "accounts")
)
``` 
Which provides valid syntax similar to this
```sql
SELECT a, table_x.b, table_q.c
FROM table_a
JOIN table_x 
    ON id=table_a_id
JOIN table_q 
    ON id=table_a_id
 WHERE id="123456" and dept="hr" or dept="accounts"
```
Even if the first code example did produce a query like the second sql output, the code itself implies that the 'join' wheres are associated with the joins themselves which is not actually the case when the query is built.  

I'm not a sql guru but I have never seen WHERE used within a JOIN (unless it is in a sub-select query as part of the ON) which is what your trying to do in this case.  

To clarify further the ON clause of a JOIN acts in the same way a WHERE does for a SELECT.

If you can provide some syntactically correct examples to show how a where is used WITHIN an ON clause I will happily stand corrected.

Cheers for the discussion ;)

circulon:
> and even if we do remove those methods thats actually a breaking change

FWIW yes it is a breaking change, but the current implementation builds invalid syntax anyway (unless a single join with a where and no other where clauses are used) so is it not worth the price for accuracy?

circulon:
Hey @josephmancuso 

I just had a thought about keeping the on_where in place.

The on_where could be adjusted to be an alias of on_value and marked as deprecated.
Possibly with a warning to use the on_value methods instead.
This would fix the invalid syntax issue while allowing users of the ORM to transition to the on_value method easily.

Admittedly this would change the output of the fully built query slightly as per the StackOverflow post I noted previously, But it would also allow a more seemless transition to using the correct methods with Join clauses.

WDYT?

josephmancuso:
Yeah I agree. It's a bit of a tough situation lol. 

On one hand using a where clause on the join and a where clause on the outer SQL creates an invalid SQL. In this case a breaking change is justified. 

On the other hand a where clause on the join WITHOUT a where clause on the outer SQL creates valid SQL which means in this case it would be a breaking change. 

I like the idea of the deprecation

circulon:
@josephmancuso 

Ok as the deprecation idea sounds like the way forward I will add the 'on_where' clauses back in as aliases for the on_value methods.   

If I get stuck with the deprecation message I will let you know.

Cheers for the feedback

josephmancuso:
@circulon Masonite has a deprecation helper you might be able to steal https://github.com/MasoniteFramework/masonite/blob/383cf4607700bfa442fa6bbf33c764035eee2fee/src/masonite/helpers/misc.py#L107

Should just be used like 

```python
@deprecated("where on joins is deprecated, please use ..")
def where(..)
```

circulon:
Hey @josephmancuso 

I hadnt encountered the 'warnings' module before, thanks for pointing it out.
used your deprected() decorator and it works like a charm.
Thanks for that.

Are there any other things we need to address?

tests/postgres/grammar/test_select_grammar.py
```diff
@@ -303,26 +303,26 @@ def can_compile_join_clause(self):
         """
         return """SELECT * FROM "users" INNER JOIN "report_groups" AS "rg" ON "bgt"."fund" = "rg"."fund" AND "bgt"."dept" = "rg"."dept" AND "bgt"."acct" = "rg"."acct" AND "bgt"."sub" = "rg"."sub\""""
 
-    def can_compile_join_clause_with_where(self):
+    def can_compile_join_clause_with_value(self):
         """
         builder = self.get_builder()
         builder.where("age", "not like", "%name%").to_sql()
```
josephmancuso: Can we just get these doctoring updated? They should just be the code we are using to generate the query.

Looks like they were wrong when we originally wrote the test but since we're in here already now it might be good to update these now

tests/sqlite/grammar/test_sqlite_select_grammar.py
```diff
@@ -303,26 +303,26 @@ def can_compile_join_clause(self):
         """
         return """SELECT * FROM "users" INNER JOIN "report_groups" AS "rg" ON "bgt"."fund" = "rg"."fund" AND "bgt"."dept" = "rg"."dept" AND "bgt"."acct" = "rg"."acct" AND "bgt"."sub" = "rg"."sub\""""
 
-    def can_compile_join_clause_with_where(self):
+    def can_compile_join_clause_with_value(self):
         """
         builder = self.get_builder()
         builder.where("age", "not like", "%name%").to_sql()
```
josephmancuso: Same with these

josephmancuso (changes requested):

josephmancuso:
Everything else looks great. I just want to do a manual test before I merge this in. Code review part looks great! good job!!

tests/postgres/grammar/test_select_grammar.py
circulon: Aaaah yes I completely overlooked that ;)
Updated now

tests/sqlite/grammar/test_sqlite_select_grammar.py
circulon: All Done

circulon:
@josephmancuso 

How is the manual checks/test coming along?

circulon:
@josephmancuso 

Bump

circulon:
@josephmancuso 
Any idea on when we can merge this in please?

I have another PR ready to go which uses this as a base and would like to get that in soon too.

Cheers

josephmancuso (approved):

josephmancuso:
Merged

PR #496, opened 2021-08-19
@BelongsToMany getting owner table name from wrong source?

garzola:
I've been trying to get the `@BelongsToMany` to work for a couple of hours now.  I kept getting an error about the model class: `AttributeError: class model 'Client' has no attribute builder`.  

When I tracked this down to masoniteorm/relationships/BelongsToMany.py, it appears that the code to get the name of the owner table is wrong.  When I fix this line (as noted in the commit), everything works as documented.

I believe this is the correct way to fix this issue, but as I have discovered on a previous commit, I may not have understood how things work enough.


josephmancuso (approved):
This should work fine. I've tested this stuff manually a lot though I'm really surprised this didn't work ...

PR #499, opened 2021-09-09
postgres use UUID datatype

circulon:
Use the UUID datatype when creating schema for primary keys and other id's
This provides schema level checking of uuid compliance.

Included an option with the .default() to use the provided string as raw sql, this allows grammar specific function/sql to (eg in postgres you can use uuid_generate_v4() ) generate a default value.


src/masoniteorm/schema/platforms/PostgresPlatform.py
```diff
@@ -56,6 +57,7 @@ class PostgresPlatform(Platform):
         "current": " DEFAULT CURRENT_TIMESTAMP",
         "now": " DEFAULT NOW()",
         "null": " DEFAULT NULL",
+        "auto_uuid": " DEFAULT uuid_generate_v4()"
```
josephmancuso: This needs a test. This is also done through the mixin.

If we are adding this as a valid default we'll need to do the same thing for the other schema classes (mysql, mssql, sqlite)

josephmancuso (changes requested):

src/masoniteorm/schema/platforms/PostgresPlatform.py
josephmancuso: Although it might be handled gracefully already but either way it needs a test to make sure adding a column default of uuid doesn't break anything

circulon: Yeah I see what you mean.  
The reson I did it like this is the DEFAULT clause is fully escaped ie quoted so the uuid_generate_v4() function cannot be added by the .default() method.

I did not find where the self.default is added to the query as I was hoping to add a .default_raw() which would be used in the same way as .default() but not quoted.
Or adding a parameter to the .default() of raw=False would also work and provide maximum flexibility.

@josephmancuso  WDYT

josephmancuso: The default values in the dictionary keys are added via the blueprint class. Then the defaults are fetched via that dictionary by the key which adds the dictionary value to the create/alter query

Marlysson:
@circulon Moving the uuid generation to database functions, how can we generate a uuid field in SQLite database?

circulon:
> @circulon Moving the uuid generation to database functions, how can we generate a uuid field in SQLite database?

@Marlysson 
This is not about moving the uuid generation to the database.  
I was looking at a grammar specific (postgres) way of generating UUID outside of the ORM scope.
ie if inserting bulk entries via csv import or some other method you need the pk to be generated at row creation.
This works in concert with the ORM UUID primary key mixin to handle (via the ORM) uuid creation   
 for grammars like SQLite that have no auto UUID generation functions (based on my limied knowledge and research of SQLite)

So I changed this PR to allow the default value as a raw sql which allows functions like NOW() for timestamps to be used incase the premapped defaults (which are quite generic) dont cover your needs.

With regards to generating a UUID specifically for SQLite PK you could use som,ething like the functions provided in this post
https://stackoverflow.com/questions/66625085/sqlite-generate-guid-uuid-on-select-into-statement
as a stored procedure then trigger it utilise it via 
```python
table.uuid("id").default("my_uuid_generator()", raw=True).primary()
```
NOTE: param raw=True is part of this PR and is subject to this PR being approved but you get the idea.

Hope this helps

src/masoniteorm/schema/Column.py
```diff
@@ -61,16 +62,18 @@ def after(self, after):
         self.after = after
         return self
 
-    def default(self, value):
+    def default(self, value, raw=False):
```
josephmancuso, line 65: So this is used like:

```python
table.string("name").default("something", raw=True)
```

?

Can we make a new method like:

```python
table.string("name").default_raw("something")
```

This would just call the `default('something', raw=True)` under the hood

josephmancuso (changes requested):

src/masoniteorm/schema/Column.py
josephmancuso: Also can we add a test for this schema? Can do it inside one of the test_postgres_alter_schema.py files

circulon: Yep NP 
Will work on these this afternoon

circulon: @josephmancuso 

Added the default_raw() method as suggested
Added variations across all grammars tests  for both .default("value", True) and .default_raw("value")

josephmancuso (dismissed):
Looks good to me.

josephmancuso:
Can you fix the conflicts here?

circulon:
> Can you fix the conflicts here?

Done and Done

PR #500, opened 2021-09-12
Fixed right join being built as left join for sqlite

circulon:


josephmancuso:
SQLite doesn't have right joins.

circulon:
> SQLite doesn't have right joins.


Aaah ok thanks @josephmancuso.  
I came across this when building the tests for a different PR.

I just learnt something new ;) 
Cheers

PR #502, opened 2021-09-15
Fix other joins cannot use lambda

circulon:
This PR enables lambda usage for left and right joins


josephmancuso (approved):

josephmancuso:
looks great! great job!

PR #505, opened 2021-09-27
Wrap all migration identifiers

circulon:
This PR fixes issues where table and column identifiers which are not wrapped (for the grammar) would be incorrectly interpreted as keywords (eg USER)


src/masoniteorm/schema/platforms/MSSQLPlatform.py
```diff
@@ -142,12 +142,12 @@ def compile_alter_sql(self, table):
                 sql.append(
                     f"ALTER TABLE {self.wrap_table(table.name)} ADD "
                     + self.get_foreign_key_constraint_string().format(
-                        clean_column=column,
+                        clean_column=self.wrap_column(column),
```
josephmancuso: I don't think these clean_columns should be wrapped I thought there was a reason I specified the clean column here. Clean meaning it is just the column value and not wrapped. Which is why we have a `column` in the same format method here

src/masoniteorm/schema/platforms/Platform.py
```diff
@@ -64,12 +64,12 @@ def foreign_key_constraintize(self, table, foreign_keys):
                 cascade += f" ON UPDATE {self.foreign_key_actions.get(foreign_key.update_action.lower())}"
             sql.append(
                 self.get_foreign_key_constraint_string().format(
-                    column=foreign_key.column,
-                    clean_column=foreign_key.column,
+                    column=self.wrap_column(foreign_key.column),
+                    clean_column=self.wrap_column(foreign_key.column),
```
josephmancuso: Same here

josephmancuso (changes requested):

src/masoniteorm/schema/platforms/MSSQLPlatform.py
circulon: @josephmancuso 
Hmm interesting...

As it turns out the clean_column specifier is not actually used in any of the platforms get_foreign_key_constraint_string() method.
looks like an oversight that crept in, so I haave removed it from both the locations it appears.

I Missed that when I was going through all of the tests.

josephmancuso (approved):

PR #506, opened 2021-09-28
Fix migration timestamps() not using datetime

circulon:
This PR aligns the created_at and updated_at columns to use the same datatype as the deleted_at column.

It also adjusts the postgres datetime datatype to use the TIMESTAMPTZ which includes timezone data


josephmancuso:
@circulon conflicts here

circulon:
> @circulon conflicts here

Done

PR #507, opened 2021-10-07
Add a way to customize configuration location

### LINKED ISSUE #501: Allow database configuration/settings path to be specified for CLI commands
It should be possible to specify where to load settings from when running migrations, seeders, etc using the `masonite-orm` CLI

This would allow the ORM to be used/integrated in existing projects without having to rewrite the rest of the application to use the predefined configuration directory (`config/database(.py)` to avoid duplication of config/settings directories.

There are multiple ways to implement the loading itself but the simplest way might be adding another CLI option that allows you to specify the configuration directory.

Example command: `masonite-orm migrate --directory=src/database/migrations --config=src/settings/database`

**Additional context**
Example class/import triggered by `masonite-orm migrate` that does not currently work if settings are not in the exact specified pypath and format.
image

This will currently result in the following module import error
```bash
(venv) C:\Users\{USER}\Code\{PROJECT}\{REPO}>masonite-orm migrate -d=src\api\database

  ModuleNotFoundError

  No module named 'config'

  at ~\Code\EVA_AST\ast_new\venv\lib\site-packages\masoniteorm\migrations\Migration.py:27 in __init__
       23│         self.migration_directory = migration_directory
       24│         self.last_migrations_ran = []
       25│         self.command_class = command_class
       26│
    →  27│         from config.database import DB
       28│
       29│         DATABASES = DB.get_connection_details()
       30│
       31│         self.schema = Schema(
```

This is already available in orator
image

~~- [ ] Is this a breaking change?~~
**Only if the existing import/path is not set as default**

girardinsamuel:
Try to fix #501.
Based on the discussions in the issue, here is what I propose:

## Normal usage
1. By default Masonite ORM will look for a `DB_CONFIG_PATH` env variable which should contain the path to the configuration file.
2. If this variable is not defined, Masonite ORM will look for a `config/database.py` file (which is the database configuration file for Masonite).

## ORM commands usage
In the commands we now have a way to override the configuration path with the `--config` option. This option, if given takes precedence over the two others options.

```bash
$ orm migrate -C app/config/db
```

So in this case it becomes:
1. Masonite ORM will use the path given in command option.
2. If no option is given, Masonite ORM will look for a `DB_CONFIG_PATH` env variable which should contain the path to the configuration file.
3. If this variable is not defined, Masonite ORM will look for a `config/database.py` file (which is the database configuration file for Masonite).

## Path to the configuration file
The path can be a python path or a file path. All paths below are accepted:
```python
app/config/database.py
app/config/database
app.config.database
app.config.database.py
```
image

Only little detail is that as : `-c` is already taken by connection and `-s` by show, I can't use them for short name of `--config` or `--settings`. So for now I have kept `--config` and `-C` (uppercased) as short name...I am open to proposals.

### DIFF
--- a/src/masoniteorm/commands/CanOverrideConfig.py
+++ b/src/masoniteorm/commands/CanOverrideConfig.py
@@ -8,6 +8,9 @@ class CanOverrideConfig(Command):
 
     def add_option(self):
         self._config.add_option(
-            "option", "o", 8, "The location of the config module", None
+            "option",
+            "o",
+            8,
+            "The path to the ORM configuration file. If not given DB_CONFIG_PATH env variable will be used and finally 'config.database'.",
+            None,
         )
-
--- a/src/masoniteorm/commands/CanOverrideConfig.py
+++ b/src/masoniteorm/commands/CanOverrideConfig.py
@@ -7,8 +7,10 @@ class CanOverrideConfig(Command):
         self.add_option()
 
     def add_option(self):
+        # 8 is the required flag constant in cleo
         self._config.add_option(
             "config",
             "C",
+            8,
             description="The path to the ORM configuration file. If not given DB_CONFIG_PATH env variable will be used and finally 'config.database'.",
         )
--- a/src/masoniteorm/commands/CanOverrideConfig.py
+++ b/src/masoniteorm/commands/CanOverrideConfig.py
@@ -1,0 +1,13 @@
+from cleo.commands.command import Command
+
+
+class CanOverrideConfig(Command):
+    def __init__(self):
+        super().__init__()
+        self.add_option()
+
+    def add_option(self):
+        self._config.add_option(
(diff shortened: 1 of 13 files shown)


girardinsamuel:
Maybe we could add an other source for the config file such as `.orm` file located at project root, especially useful if you want to use Masonite ORM standalone without declaring environment variables.
Such a project would be like this:

```
# layout
your_program.py
config.py
.orm
```

config.py
```
#... usual masonite orm config file
```

.orm
```
config
```

Note that the config file is not even required you could have:
```
program.py
.orm
```

program.py could be:
```python
# ORM config 
DATABASES = { ... }
DB = ConnectionResolver().set_connection_details(DATABASES)
# ....

# your program code
# ...
```
and .orm would contain
```
program
```

josephmancuso:
merging into 2.0 branch

PR #509, opened 2021-10-25
Fix Faker being reset for every registered call and after create

circulon:
This Fixes the faker instance producing duplicate data when using .unique
This was due to a new Faker instance being created every time a factory callback is used

Example being if a factory is used to create multiple  users that each required a unique email address, the same email address would be generated every time even if faker.unique.ascii_email is used

### DIFF
--- a/src/masoniteorm/factories/Factory.py
+++ b/src/masoniteorm/factories/Factory.py
@@ -1,10 +1,21 @@
 from faker import Faker
+import random
 
 
 class Factory:
 
     _factories = {}
     _after_creates = {}
+    _faker = None
+
+    @property
+    def faker(self):
+        if not Factory._faker:
+            Factory._faker = Faker()
+            random.seed()
+            Factory._faker.seed_instance(random.randint(1, 10000))
+
+        return Factory._faker
 
     def __init__(self, model, number=1):
         self.model = model
@@ -15,7 +26,7 @@ class Factory:
             dictionary = {}
 
         if self.number == 1 and not isinstance(dictionary, list):
-            called = self._factories[self.model][name](Faker())
+            called = self._factories[self.model][name](self.faker)
             called.update(dictionary)
             model = self.model.hydrate(called)
             self.run_after_creates(model)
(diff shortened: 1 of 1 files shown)


josephmancuso:
why don't you want a new instance of Faker? Whats the benefit of having a singleton style faker instance?

circulon:
> why don't you want a new instance of Faker? Whats the benefit of having a singleton style faker instance?

Ok lets take this scenario:
a company has a unique email address (email has a UNIQUE Index on the column)
Each user in that company also has a unique company email address (email has a UNIQUE Index on the column).
This is a fairly standard data scenario.

For a Faker instance to generate unique data it MUST be seeded, which it is not by default, as such it will generate the same data for .unique calls. 

In its current state the following factories will fail with a duplicate email error on the second user of the first company as the same name is generated for the new instance of faker applied to each user model in after_create.

```python
def company_factory(faker):
    return {
        "name": faker.unique.company(),
        "email": faker.unique.ascii_company_email(),
        "phone": faker.unique.phone_number(),
    }

def user_factory(faker):
    return {}

def after_company(model, faker):
    company_email = model.email.split(sep="@")

    user_count = faker.random_int(min=2, max=25)
    for i in range(user_count):
        user_name = faker.unique.name().split()
        user_email = "{}.{}@{}".format(user_name[0], user_name[1], company_email[1]).lower().replace("..", ".")
        result = User.create({
            "email": user_email,
            'company_id': model.id,
            "first_name": user_name[0],
            "last_name": user_name[1],
            "phone": faker.unique.phone_number(),
           })

Factory.register(Company, company_factory)
Factory.register(User, user_factory)
Factory.after_creating(Company, after_company)

Factory(Company, 10).create()

```

By using the Faker instance in a singletome type manner the .unique(ness) of all values is ensured across the entire seed:run

@josephmancuso 
I hope this clarifies things

josephmancuso:
@circulon got it. That makes a lot of sense actually

josephmancuso (approved):

circulon:
Hi @josephmancuso 

Would it be possible to get these changes merged and a new release tagged please?

josephmancuso:
@circulon sure I'll make sure I bump priority on this

josephmancuso:
@circulon this is in 1.0.57

PR #518, opened 2021-11-12
fixed datetimes not being converted in models

### LINKED ISSUE #513: Creating or updating records with pendulum instance does not work
**Describe the bug**
When creating or updating ORM records with Model.create() or model.update() or by directly setting attribute does not work with SQLite database.

**Expected behavior**
As we have declared the datetime field in `__dates__` it is expected that Masonite ORM will serialize date/datetime objects from pendulum instance before inserting records into db.

image

 - OS: MacOS
 - Version Big Sur

 - Type: SQLite
 - Masonite ORM 2.0

josephmancuso:
Closes #513 
@girardinsamuel This should work right?

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -637,6 +637,9 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
         if attribute in self.__casts__:
             value = self._set_cast_attribute(attribute, value)
 
+        if attribute in self.get_dates():
+            value = self.get_new_datetime_string(value)
+
         try:
             if not attribute.startswith("_"):
                 self.__dict__["__dirty_attributes__"].update({attribute: value})


girardinsamuel:
Maybe but I am not sure, where this method is called. Maybe it would be ideal to write a unit test to cover those use cases:
```python
import pendulum
now = pendulum.now()
# use case 1
User.create({"name": "John", "verified_at": now})
# use case 2
user = User.find(1)
user.update({"verified_at": now})
# use case 3
user = User.find(1)
user.verified_at = now
user.save()
```

girardinsamuel (changes requested):
Maybe... but I am not sure where this method is called. Maybe it would be ideal to write a unit test to cover those use cases:
```python
import pendulum
now = pendulum.now()
# use case 1
User.create({"name": "John", "verified_at": now})
# use case 2
user = User.find(1)
user.update({"verified_at": now})
# use case 3
user = User.find(1)
user.verified_at = now
user.save()
```

PR #521, opened 2021-11-22
Feat/516 - Migrate and Rollback migration by name

### LINKED ISSUE #516: Migrate and rollback by name
I frequently have multiple migrations for various parts of my project in development at once. 

At the moment, it is awkward to avoid running all of them because craft migrate runs all that are not yet applied by default. I typically move the undesired or incomplete migration files out of the directory, run craft migrate to do the ones I want, and then put the others back to continue development.

It would be nice if I could pass a filename to the migrate command and it would run that one file (if it hasn't already been run). Similarly, craft migrate:rollback should allow rollback by name instead of the entire batch.

(Additionally helpful and somewhat related would be adding a `craft migrate:status` command that shows which have been run or not. I was thinking this already existed or may be shown as a side effect of running other commands, but perhaps I'm thinking of the ruby on rails equivalents.)

Marlysson:
Closes #516.

Add option to define migration to handle with. Migrate or rollback:

```
masonite-orm migrate --m migration_name
masonite-orm migrate:rollback --m migration_name
```

Ps: the value to migration name "migration_name" will be given by the results listed by migrate:status command.

- [ ] - Add modifications in doc repository

### DIFF
--- a/src/masoniteorm/migrations/Migration.py
+++ b/src/masoniteorm/migrations/Migration.py
@@ -106,21 +106,18 @@ class Migration:
                 ran.append(migration)
         return ran
 
-    def migrate(self, output=False):
+    def migrate(self, migration=None, output=False):
+        
+        migrations_to_migrate = self.get_unran_migrations() if migration == 'all' else [migration]
+
         batch = self.get_last_batch_number() + 1
 
-        for migration in self.get_unran_migrations():
-            migration_module = migration.replace(".py", "")
-            migration_name = camelize(
-                "_".join(migration.split("_")[4:]).replace(".py", "")
-            )
-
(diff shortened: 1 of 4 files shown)


src/masoniteorm/migrations/Migration.py
```diff
@@ -106,21 +106,18 @@ def get_ran_migrations(self):
                 ran.append(migration)
         return ran
 
-    def migrate(self, output=False):
-        batch = self.get_last_batch_number() + 1
+    def migrate(self, migration=None, output=False):
```
josephmancuso: this should be `migration=all`

josephmancuso (changes requested):

Issue #494, opened 2021-08-13
Sub Selects binding error for multiple where clauses


circulon:
**Describe the bug**
Using a sub select that has more than 1 filter in the where clause throws a binding error (too many arguments) when using .get().

Strangely the to_sql() does not throw this error (when using the same query) and returns a properly built query that is testable in db clients 

**To Reproduce**
```python
test_table = SomeModel.get_table_name()
date_today = date_utc.add(days=1).to_date_string()
date_7_days_ago = date_utc.subtract(days=6).to_date_string()
stats_builder = (
            QueryBuilder()
            .from_(test_table)
            .where("user_id", current_user.id)
            .count("id as all_time_count")
            .avg("rating as all_timerating_avg")
            .max("rating as all_time_rating_max")
            .add_select(
                "count_days_7",
                lambda query: (
                    query.count("id as lc".format(test_table))
                    .from_(test_table)
                    .where("user_id", current_user.id)
                    # adding 1 or more of the following results in a binding error
                    # Yes I know they are the same but this is just an example ;)
                    .between('updated_at', date_7_days_ago, date_today)  
                    .where("updated_at", ">=", date_7_days_ago)
                    .where("updated_at", "<=", date_today)
                   
                   # Where_raw DOES NOT throw the binding error presumably because 
                   # the values are inserted into the string and not being quoted as per bindings are 
                   .where_raw(
                        "updated_at BETWEEN '{}' AND '{}'".format(
                            date_7_days_ago, date_today
                        )
                    )
                ),
            )

# this works
logging.info(stats_builder.to_sql())
# this throws a binding error        
stats_result = stats_builder.get()[0]
```

**Expected behavior**
Sub selects with multiple where filters should work as expected

**Desktop (please complete the following information):**
 - OS: Mac OSX
 - Version 10.15.7

**What database are you using?**
 - Type: Postgres 
 - Version 10.7
 - Masonite ORM v1.0.55

josephmancuso:
Fixed in #542

PR #542, opened 2021-12-15
Fixed wrong spread bindings

felipehertzer:
I am developing an application in Python and I had the same problem related to issue #494 where it is not possible to use multiple wheres in a subquery

The add_binding function only expects a 'binding' parameter, however the value.builder._bindings variable is a list of bind values.

`add_binding(self, binding):`

`value.builder._bindings = ['value1', 'value2']`

In the current code a spread is being performed, however the list is one-dimensional this is causing the problem of too many arguments when multiple arguments in a subquery

**Current:**
```python
self.add_binding(*value.builder._bindings) --- result --> self.add_binding('value1', 'value2')
```

**New:**
```python
self.add_binding(value.builder._bindings) --- result --> self.add_binding(['value1', 'value2'])
```

### DIFF
--- a/src/masoniteorm/query/grammars/BaseGrammar.py
+++ b/src/masoniteorm/query/grammars/BaseGrammar.py
@@ -636,7 +636,7 @@ class BaseGrammar:
                 if qmark:
                     query_from_builder = value.builder.to_qmark()
                     if value.builder._bindings:
-                        self.add_binding(*value.builder._bindings)
+                        self.add_binding(value.builder._bindings)
                 else:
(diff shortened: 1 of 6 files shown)


Marlysson (dismissed):

josephmancuso:
This just needs a test before I merge this in. Code looks good but I like to write tests for bugs like this

felipehertzer:
Hello @josephmancuso thank you for reply. I just added the tests.

josephmancuso (approved):

josephmancuso:
well done 👏 Great quality bug fix

PR #552, opened 2021-12-18
made some changes for serialize

erhuabushuo:
1. add include argument to serialize,  `Explicit is better than implicit`. 
2. when we pass optional argument `exclude`, just override default __hidden__, 
3. pagination also need pass optional arguments to serialize
4. do not use mutable objects`list` as default arguments [https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/](url)

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -449,7 +449,7 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
             .first()
         )
 
-    def serialize(self, exclude=[]):
+    def serialize(self, exclude=None, include=None):
         """Takes the data as a model and converts it into a dictionary.
 
         Returns:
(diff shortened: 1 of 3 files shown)


src/masoniteorm/models/Model.py
```diff
@@ -449,15 +449,25 @@ def fresh(self):
...
         Returns:
             dict
         """
         serialized_dictionary = self.__attributes__
 
-        self.__hidden__ += exclude
+        # prevent using both exclude and include at the same time
+        if exclude is not None and include is not None:
+            raise AttributeError(
+                f"defines both exclude and include paramter."
```
josephmancuso: `paramter ` -> parameter

josephmancuso: Actually change to "Cannot define both includes and exclude values"

josephmancuso (changes requested):

josephmancuso (approved):

PR #555, opened 2021-12-21
Add ability to override default command options values from init()

girardinsamuel:
Now for all ORM commands we can update values of default options when initializing the command:

### Example with MakeModelCommand:

Default for `directory` is:
```python
```
You need to pass a keyword argument with the same name as the option:
```python
MakeModelCommand(directory="app/models")
```
And 🎉 :
image

This will be useful to edit options from Masonite.

### DIFF
--- a/src/masoniteorm/commands/MakeModelCommand.py
+++ b/src/masoniteorm/commands/MakeModelCommand.py
@@ -3,9 +3,10 @@ import pathlib
 
 from cleo import Command
 from inflection import camelize, underscore, tableize
+from .CanOverrideOptionsDefault import CanOverrideOptionsDefault
 
 
-class MakeModelCommand(Command):
+class MakeModelCommand(CanOverrideOptionsDefault, Command):
     """
     Creates a new model file.
 
--- a/src/masoniteorm/commands/MakeModelCommand.py
+++ b/src/masoniteorm/commands/MakeModelCommand.py
@@ -1,12 +1,12 @@
 import os
 import pathlib
 
-from cleo import Command
-from inflection import camelize, underscore, tableize
-from .CanOverrideOptionsDefault import CanOverrideOptionsDefault
+from inflection import camelize, tableize
+
+from .Command import Command
 
 
-class MakeModelCommand(CanOverrideOptionsDefault, Command):
+class MakeModelCommand(Command):
     """
     Creates a new model file.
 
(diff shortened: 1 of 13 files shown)


girardinsamuel:
@josephmancuso Is it okay for you ?

josephmancuso (approved):
great solution!

girardinsamuel:
Or you do you want that we subclass all Command from a local Command so that all future commands added in the ORM implements this mixin.. ?

```python
# Command.py
from cleo import Command as BaseCommand
from .CanOverrideOptionsDefault import CanOverrideOptionsDefault

class BaseCommand(CanOverrideOptionsDefault, BaseCommand):
    pass
```
And then:
```python
# MakeModelCommand.py
from .Command import Command

class MakeModelCommand(Command):

     pass
```

josephmancuso:
> Or you do you want that we subclass all Command from a local Command so that all future commands added in the ORM implements this mixin.. ?
> 
> ```python
> # Command.py
> from cleo import Command as BaseCommand
> from .CanOverrideOptionsDefault import CanOverrideOptionsDefault
> 
> class BaseCommand(CanOverrideOptionsDefault, BaseCommand):
>     pass
> ```
> 
> And then:
> 
> ```python
> # MakeModelCommand.py
> from .Command import Command
> 
> class MakeModelCommand(Command):
> 
>      pass
> ```

This is an even better idea lol

src/masoniteorm/commands/MakeModelCommand.py
```diff
@@ -1,8 +1,9 @@
 import os
 import pathlib
 
-from cleo import Command
-from inflection import camelize, underscore, tableize
+from inflection import camelize, tableize
```
girardinsamuel, line 4: underscore was not used

src/masoniteorm/commands/MakeSeedCommand.py
```diff
@@ -1,10 +1,9 @@
 import os
 import pathlib
 
-from cleo import Command
-from inflection import camelize
+from inflection import camelize, underscore
 
-from inflection import underscore
+from .Command import Command
```
girardinsamuel, line 6: underscore was not used

girardinsamuel:
@josephmancuso  ready !

josephmancuso (approved):

PR #556, opened 2021-12-21
WIP: Add missing argument options to some commands

girardinsamuel:
- MakeObserver was missing `directory` option
- MakeModel was missing `migrations-directory` option when the migration file is created with it
- MigrateRefresh was missing `seeds-directory` option when seed are ran after the command

### DIFF
--- a/src/masoniteorm/commands/MakeModelCommand.py
+++ b/src/masoniteorm/commands/MakeModelCommand.py
@@ -16,6 +16,7 @@ class MakeModelCommand(Command):
         {--c|create : If the migration file should create a table}
         {--t|table : If the migration file should modify an existing table}
         {--d|directory=app : The location of the model directory}
+        {--D|migrations-directory=databases/migrations : The location of the migration directory}
     """
 
     def handle(self):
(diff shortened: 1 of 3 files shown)


girardinsamuel:
> @josephmancuso an other small fix on commands (something I saw when working on integrating commands in M4). ready for review.

Actually, wait I'll need to check that I can parse argument with a dash in the middle ...

src/masoniteorm/commands/MakeModelCommand.py
```diff
@@ -16,6 +16,7 @@ class MakeModelCommand(Command):
         {--c|create : If the migration file should create a table}
         {--t|table : If the migration file should modify an existing table}
         {--d|directory=app : The location of the model directory}
+        {--D|migrations-directory=databases/migrations : The location of the migration directory}
```
josephmancuso, line 19: This is the same as -d right? what is -D?

josephmancuso: never I see lol

josephmancuso (approved):

src/masoniteorm/commands/MakeModelCommand.py
girardinsamuel: The issue with Cleo commands is that the short name of the option needs to be only one letter... I cannot use `md` for migrations directory...so either you have a better idea for the letter, either I went with the uppercased version.

girardinsamuel: It's already the case with the config option which is using `-C` because `-c` is taken by connection option

josephmancuso: No worries I thought you were doing the model directory but I read "model directory" instead of "migration directory"

PR #557, opened 2021-12-21
Fix overriding command options when using dash in option name

girardinsamuel:
```python
# for this command
        {--D|migrations-directory=databases/migrations : The location of the migration directory}

# you would override it like this
MakeModelCommand(app, directory="app/models", migrations_directory="app/db/migrations")
```


girardinsamuel:
@josephmancuso a small fix that I did not have time to put in last PR. You're reviewing and merging like ⚡ !

josephmancuso (approved):

josephmancuso:
lol trying to keep you 🔥

PR #558, opened 2021-12-22
Add new `shell` command to connect to database shell programs

girardinsamuel:
- Run `masonite-orm shell` to connect to database shell program of your default connection.
- Run `masonite-orm shell -c mysql` to connect to database shell program of the given connection
- Run `masonite-orm shell -s` to connect and display the command that will be called to connect.

image

image

image

when `-s` option is used any sensitive data will be hidden to avoid being displayed in the term (and eventually logged in bash history...or somewhere else).
image

### DIFF
--- a/src/masoniteorm/commands/ShellCommand.py
+++ b/src/masoniteorm/commands/ShellCommand.py
@@ -1,0 +1,151 @@
+import subprocess
+import shlex
+from collections import OrderedDict
+
+from ..config import load_config
+
+from .Command import Command
+
+
+class ShellCommand(Command):
+    """
+    Connect to your database interactive terminal.
+
+    shell
+        {--c|connection=default : The connection you want to use to connect to interactive terminal}
+        {--s|show=? : Display the command which will be called to connect}
+    """
+
+    shell_programs = {
+        "mysql": "mysql",
+        "postgres": "psql",
+        "sqlite": "sqlite3",
+        "mssql": "sqlcmd",
+    }
+
+    def handle(self):
+        resolver = load_config(self.option("config")).DB
+        connection = self.option("connection")
+        if connection == "default":
+            connection = resolver.get_connection_details()["default"]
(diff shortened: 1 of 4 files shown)


josephmancuso:
This is an amazing idea!

girardinsamuel:
> This is an amazing idea!

Let me know, if something is not clear. I tested it manually for all drivers excepted for MSSQL as I don't have this configured...

girardinsamuel:
Also added unit tests here ✅ which has allowed me to find and fix some bugs 💪.

josephmancuso:
I couldn't test this because I don't have shells setup but ill merge this anyway

girardinsamuel:
> I don't have shells setup

What do you mean by this `I don't have shells setup` ?

josephmancuso:
like I can't run `mysql` in my terminal to start up a mysql shell

src/masoniteorm/commands/ShellCommand.py
```diff
@@ -0,0 +1,198 @@
...
+            self.comment(cleaned_command)
+            self.line("")
+
+        # let shlex split command in a list as it's safer
+        command_args = shlex.split(command)
+        try:
+            subprocess.run(command_args, check=True, env=env)
+        except FileNotFoundError:
+            self.line(
+                f"<error>Cannot find {config.get('full_details').get('driver')} program ! Please ensure you can call this program in your shell first.</error>"
```
josephmancuso, line 53: I was getting this error for every connection but I guess because I didn't setup a mysql shell program

girardinsamuel: oh okay, but it's weird, I think that when I install mysql or postgres server, those shell programs were bundled. I did not installed them separately. I installed those with Homebrew.

girardinsamuel: Anyway in the meantime you can test with sqlite.
SQLite has a shell !

PR #559, opened 2021-12-22
Add support for `DATABASE_URL`

### LINKED ISSUE #307: Support database URLs
We should be able to somehow support database URL's 

postgres://user:secret@localhost:5432/mydatabasename

We should be able to discuss the best way to approach this. Maybe we either have a function that returns the information in a correct format like:

```python
# config/database.py
from masoniteorm.helpers import database_url

DATABASES = {
    'mysql': database_url(env("DATABASE_URL"))
}
```

Maybe this returns a dictionary.

Another option is maybe just adding a URL attribute, and if it exists, use that instead of all the other config options:

```python
# config/database.py
from masoniteorm.helpers import database_url

DATABASES = {
    'mysql': {
        "user": "..",
        "password": "..",
        "url": "mysql://user:secret@localhost:5432/mydatabasename"
    }
}
```

If `url` is set, use that, if not then don't use it? 

Open for discussion. Can see how other libraries do this

girardinsamuel:
Fix #307.

Here is a proposal for supporting `DATABASE_URL`:

Import the helper
```python
from masoniteorm.config import db_url
```

This helper will return a full config dict that you can directly use in `DATABASES` config:

```python
DATABASES = {
    "default": "mysql",
    "mysql": db_url(),
}
```
Here as no url string is given it will automatically use `DATABASE_URL` env (as often exposed by PaaS service: Heroku...)

You want to use an other env variable ? No problem:
```python
DATABASES = {
    "default": "mysql",
    "mysql": db_url(os.getenv("DB_URL")),
}
```

Of course you can specify an url directly:
```python
DATABASES = {
    "default": "mysql",
    "mysql": db_url("mysql://root:@localhost:3306/orm"),
}
```

...and you can specify additional parameters:
```python
DATABASES = {
    "default": "mysql",
    "mysql": db_url("mysql://root:@localhost:3306/orm", log_queries=True, prefix="", options={}),
}
```

...or you can just merge it with some other options (not related to the url):
```python
DATABASES = {
    "default": "mysql",
    "mysql": {
        **db_url("mysql://root:@localhost:3306/orm"),
        "prefix": "",
        "options": {},
        log_queries=True
     }
}
```

It support all drivers.

For SQLite it's:
```python
db_url("sqlite://orm.sqlite3")
```

🎉 🚀 

If it's okay, then I could add some unit tests.


config/database.py
```diff
@@ -1,11 +1,11 @@
 """ Database Settings """
-
 import os
+import logging
```
girardinsamuel, line 3: I just cleaned and re-organized the imports...

josephmancuso: Do we need this file? I thought the database.py file is inside `tests/integrations/config/databse.py`?

src/masoniteorm/config.py
```diff
@@ -21,3 +22,95 @@ def load_config(config_path=None):
...
+        "mysqlgis": "mysql",
+        "mysql-connector": "mysql",
+        "mssql": "mssql",
+        "sqlite": "sqlite",
+    }
+
+    # this is a special case, because if we pass this URL into
+    # urlparse, urlparse will choke trying to interpret "memory"
+    # as a port number
+    if url == "sqlite://:memory:":
```
josephmancuso: We could probably check for a couple of string here. Maybe also just `:memory:` too

config/database.py
girardinsamuel: in Masonite yes...but in standalone ORM here we need this file or at least a file defined by DB_CONFIG_PATH. So I guess we keep it for this repo but it's indeed the file tests/integrations/config/database.py for masonite repo.

src/masoniteorm/config.py
josephmancuso: Nobody is going to remember this string without referring to the docs

josephmancuso (commented):
Looks great! Great implementation

src/masoniteorm/config.py
girardinsamuel: Yeah we could check just `:memory:`.  What do you mean by couple of strings lol ?

config/database.py
josephmancuso: So do we just need to set the env in the GitHub actions file?

src/masoniteorm/config.py
josephmancuso:
```python
if url in ("sqlite://:memory:", ":memory:")
```

I am also assuming that `sqlite://:memory:` is the officially supported string. I just want to be able to remember a string and I'm not going to remember `sqlite://:memory:` lol

josephmancuso: I'll forget one of the 3 `:` and pull my hair out why its not working

girardinsamuel: Oh okay got it. Actually you can already go the simple way`sqlite://` but I can add `sqlite://memory` also.

girardinsamuel: Now you can use those for SQLite in memory:
```python
db_url("sqlite://")
db_url("sqlite://memory")
db_url("sqlite://:memory:")  # the official one
```
And for a SQLite classic db file:
```python
db_url("sqlite://orm.sqlite3")
```

girardinsamuel:
> Looks great! Great implementation

Unit tests have been added ✅ !

josephmancuso (approved):

PR #562, opened 2021-12-26
Add fix for windows

girardinsamuel:
When loading Python paths to get classes, on Windows paths contains `//` and need to be processed in order to be converted to a module path.


girardinsamuel:
@josephmancuso This is a bug fix for Windows for 2.0. I will open another PR to fix this bug in 1.0 if needed.

josephmancuso (approved):

PR #566, opened 2022-01-05
fixes issue

josephmancuso:


josephmancuso:
This fixes the error but there is no deleted at clause where the join when there is a mixin. I don't know if it should or not honestly.

josephmancuso (commented):
approved

PR #571, opened 2022-01-15
Added create_table_if_not_exists schema

yubarajshrestha:
Adds new feature `create_table_if_not_exists` to the schema.

### DIFF
--- a/src/masoniteorm/schema/platforms/MSSQLPlatform.py
+++ b/src/masoniteorm/schema/platforms/MSSQLPlatform.py
@@ -283,7 +283,7 @@ class MSSQLPlatform(Platform):
 
     def create_format(self):
         return "CREATE TABLE {table} ({columns}{constraints}{foreign_keys})"
-    
+
     def create_if_not_exists_format(self):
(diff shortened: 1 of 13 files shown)


josephmancuso (changes requested):
Looks good but i don't see any tests for this

josephmancuso:
I'll add a test now

src/masoniteorm/schema/platforms/MSSQLPlatform.py
```diff
@@ -284,6 +284,9 @@ def get_column_string(self):
     def create_format(self):
         return "CREATE TABLE {table} ({columns}{constraints}{foreign_keys})"
 
+    def create_if_not_exists_format(self):
+        return "CREATE TABLE IF NOT EXISTS {table} ({columns}{constraints}{foreign_keys})"
```
josephmancuso, line 288: There actually is no equivalent in Microsoft SQL Server

josephmancuso (changes requested):

josephmancuso (approved):
I'll merge this in anyway. We can cross that bridge when we get there

PR #598, opened 2022-02-19
Rename wrong doc-string in observer stub

fathur:
This PR intended to fix the wrong doc-string when generating observer for specific model


girardinsamuel (approved):
Thanks for finding this out and fixing this ! Looks good to me 👍

fathur:
Thanks, @girardinsamuel. This is my first-time contribution, it seems needs maintainer @josephmancuso approval also

josephmancuso (approved):

PR #615, opened 2022-03-30
Fix/612

### LINKED ISSUE #612: MySQL foreign key referencing a big_increments field
Hi all! I am trying to create a foreign key referencing a big_increments field, but I get `(1215, 'Cannot add foreign key constraint')` error. I think it is because I can't create a field that is the same type as big_increments (unsigned non nullable auto incrementing big integer with length of 20) because when I tried creating `big integer` with `unsigned()` modifier, its length got shortened to 10.

Example of my code:
```
class CreateUsersTable(Migration):
    def up(self):
        """
        Run the migrations.
        """
            table.big_increments('id')
            ...

class CreateAppsTable(Migration):

    def up(self):
        """
        Run the migrations.
        """
        with self.schema.create('apps') as table:
            table.big_increments('id')
            table.big_increments('user_context_id', length=20).unsigned().nullable()
            ...

            table.index('user_context_id', name=apps_user_context_id_index')
            table.foreign('user_context_id', name='apps_user_context_id_foreign') \
                .references('id').on('users') \
                .on_delete('cascade')
```

Can you please check if it is possible to create foreign key  referencing a big_increments field?

 - OS: Linux
 - Version Ubuntu 21.10
 
 - Type: [MySQL]
 - Version [14.14 Distrib 5.7.30, for Linux (x86_64) using  EditLine wrapper]
 - Masonite ORM [2.4.2]

josephmancuso:
Fixes issue with unsigned types

Closes #612

### DIFF
--- a/src/masoniteorm/schema/Blueprint.py
+++ b/src/masoniteorm/schema/Blueprint.py
@@ -634,7 +634,7 @@ class Blueprint:
         _columns = []
         _columns.append(
             self.table.add_column(
-                "{}_id".format(column), "unsigned_integer", nullable=nullable
+                "{}_id".format(column), "integer_unsigned", nullable=nullable
             )
         )
         _columns.append(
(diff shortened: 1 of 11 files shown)


josephmancuso (commented):
reviewed

PR #616, opened 2022-03-30
fixed soft deletes changing columns

### LINKED ISSUE #613: SoftDeletesMixin doesn't work as expected when trying to change the column name.
SoftDeletesMixin doesn't work as expected when trying to change the column name. 

When we try to add `__deleted_at__ = 'archived_at'` to the Model class, it changes the value of __deleted_at__ field on the class but it is not used later when creating a query. I think the reason is this in masoniteorm.models.Model, method boot, [this](https://github.com/MasoniteFramework/orm/blob/99f32b18b1c5e227edb156ecd2d6205bde63e73e/src/masoniteorm/models/Model.py#L280) line:

```
if class_name.endswith("Mixin"):
       getattr(base_class(), "boot_" + class_name)(self.builder)
```
because it doesn't use previously changed mixin, but creates a new one. 

I think it can be solved like this:
```
if class_name.endswith("Mixin"):
       getattr(self, "boot_" + class_name)(self.builder)
```
but i am not sure if it will break some other mixin.

Can you please tell me if this change is safe to add, and I will create a PR if it is. If not, do you have some other solution to my problem?

Desktop:

    OS: Linux
    Version Ubuntu 21.10

What database are you using?

    Type: [MySQL]
    Version [14.14 Distrib 5.7.30, for Linux (x86_64) using EditLine wrapper]
    Masonite ORM [2.4.2]

josephmancuso:
Closes #613

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -278,7 +278,7 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
                 class_name = base_class.__name__
 
                 if class_name.endswith("Mixin"):
-                    getattr(base_class(), "boot_" + class_name)(self.builder)
+                    getattr(self, "boot_" + class_name)(self.builder)
 
             self._booted = True
             self.observe_events(self, "booted")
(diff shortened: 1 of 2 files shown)


josephmancuso (commented):
reviewed

PR #618, opened 2022-04-07
Feature/617 - Adds REGEXP and NOT REGEXP operators to query builder

### LINKED ISSUE #617: REGEXP operator support for MySQL
Hi, we would like to be able to use REGEXP when making queries on MySQL database.

Currently when we try to execute this code:
```
query.where('name', 'regexp', f'^{value}(-[0-9]+)?$') 
```
we get error:
`ValueError: Invalid comparison operator. The operator can be =, >, >=, <, <=, !=, <>, like, not like`

I think this change should be added to the QueryBuilder, Grammars (not only Mysql) and expressions (JoinClause).

josephmancuso:
Closes #617 

Also adds regexp support to SQLite


josephmancuso (commented):
reviewed

PR #622, opened 2022-04-11
Bump faker dep requirement to allow for using latest.

stoutput:
the only reason to use <v5 was to maintain support for Python 2, which Masonite doesn't support.

With the latest versions of Faker we can get some nice features such as method type-hinting.


josephmancuso (approved):

josephmancuso:
LGTM 👍

PR #623, opened 2022-04-11
Move first query builder instantiation out of init and into Model.boot()

stoutput:
This allows for logic affecting the query builder (such as db connection resource injection) to be injected by observers' "booting" events

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -216,7 +216,6 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
         self._relationships = {}
         self._global_scopes = {}
 
-        self.get_builder()
         self.boot()
 
     @classmethod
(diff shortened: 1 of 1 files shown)


josephmancuso:
hmm interesting. Tests seem to pass 👍

josephmancuso:
was there any additional manual testing you did to prove this works?

stoutput:
The docs say that the tests were written without need for a db (https://github.com/MasoniteFramework/orm/blob/2.0/CONTRIBUTING.md#running-tests), but in my experience, it wasn't working without configuring databases. There's probably some step I was missing. I have another branch I'm working on trying to get local tests passing with docker containers: https://github.com/stoutput/orm/tree/fix-local-tests. My branch was working a week or so ago but I made some changes along the way that broke it. It's almost there, and would be nice to not have to rely on a local MySQL instance.

stoutput:
I would have liked to add a test case asserting that the DB config wasn't accessed until after the "booting" event had fired, but otherwise, I wouldn't expect this to break any existing tests. We're lazily instantiating it the first time we try to access it, during the `boot` method, rather than eagerly in `__init__`.

josephmancuso:
Agreed. I just wanna take a deeper look at this tonight and make sure theres no hidden assumptions in the architecture that would break this.

josephmancuso (approved):

PR #629, opened 2022-04-18
Add `find_or_fail` on Model

girardinsamuel:
Add query builder equivalent method for Model.

This will act as `find` but will raise a `ModelNotFound` exception if no record exists in db.

This will allow to handle those exceptions specifically, e.g. for Masonite framework.


girardinsamuel:
@josephmancuso ready for review !
Now we can use try to handle this exception in Masonite to return a 404.

josephmancuso (approved):

PR #632, opened 2022-04-21
Added .first_where() method on QueryBuilder

JarriqTheTechie:
Added `.first_where()` method to QueryBuilder.

This is a shortcut for `.where(....).first()` allowing this instead `.first_where(.....)`

This function is useful as it reduces the amount of repetitive code written when trying to select the first record matching a given criteria.


JarriqTheTechie:

josephmancuso (approved):

PR #636, opened 2022-04-21
Fix/635: Builder not initialized in many to many relations

### LINKED ISSUE #635: Builder is not initialized when attaching entities in a many to many relation
**Describe the bug**
If you try to call the `attach` method to connect two model instances that are in a many to many relation, the following error is raised:
```python
AttributeError: class model 'MyModel' has no attribute builder
```

**To Reproduce**

1. Declare two models connected in a many to many relation:
   ```python
   class Ssl(Model):
       __table__ = 'ssls'
       @belongs_to_many('ssl_id', 'app_id', 'id', 'id', table='ssl_app')
       def apps(self):
           from .app import App
           return App
    
    class App(Model):
       __table__ = 'apps'
       @belongs_to_many('app_id', 'ssl_id', 'id', 'id', table='ssl_app')
       def ssls(self):
           from .ssl import Ssl
           return Ssl
   ```
2. Attach one entity to the other one:
   ```python
   ssl = Ssl.find(1)
   app = App.find(2)
   
   ssl.attach('apps', app)   # Exception!
   # or simmetrically:
   app.attach('ssls', ssl)   # Exception!
   ```
This happens because the `BelongsToMany` class `attach` method access builder without checking whether it is initialized or not: https://github.com/MasoniteFramework/orm/blob/798de75c84fa228b9fb5dcdd3fd2ff25ede369ca/src/masoniteorm/relationships/BelongsToMany.py#L366

One workaround is to call `get_builder()` before calling `attach`:
```python
ssl.get_builder()
ssl.attach('apps', app)   # now it works!
```
but I think the proper fix would be to change the `BelongsToMany` method in this way:
```python
# Pivot.on(current_model.builder.connection) 
Pivot.on(current_model.get_builder().connection)
```

**Expected behavior**
The entities should be connected in the relation without issues.

 - OS: Ubuntu
 - Version: 21.04

 - Type: MySQL
 - Version: 14.14 Distrib 5.7.30, for Linux (x86_64)
 - Masonite ORM 2.6.1

federicoparroni:
Fix #635

### DIFF
--- a/src/masoniteorm/relationships/BelongsToMany.py
+++ b/src/masoniteorm/relationships/BelongsToMany.py
@@ -363,7 +363,7 @@ class BelongsToMany(BaseRelationship):
             )
 
         return (
-            Pivot.on(current_model.builder.connection)
+            Pivot.on(current_model.get_builder().connection)
             .table(self._table)
             .without_global_scopes()
             .create(data)
@@ -380,7 +380,7 @@ class BelongsToMany(BaseRelationship):
         )
 
         return (
-            Pivot.on(current_model.builder.connection)
+            Pivot.on(current_model.get_builder().connection)
             .table(self._table)
             .without_global_scopes()
             .where(data)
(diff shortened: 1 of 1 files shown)


josephmancuso (approved):

josephmancuso:
LGTM 👍

Issue #641, opened 2022-04-23
Custom path to DB connection Details ignored


circulon:
**Describe the bug**
Using custom connection details providing a non default path to database config file are ignored.

**To Reproduce**

i ran into this while playing with a custom  Masonite config 
But this will also be the case if others use a non standard location for their config

1. setup basic Masonite App 
2. make a new module called "core" inside the "config" module
3. move all  the .py files from "config" module into "config.core" module
4. edit the Kernel.py file
5. in the register_configurations method change 
self.application.bind("config.location", "config")
to
self.application.bind("config.location", "config.core") 
6. run the local server
7. see the error
      masoniteorm.exceptions.ConfigurationNotFound: ORM configuration file has not been found in config.database. 

**Expected behavior**
the custom config path should be used to load the config file 

**Screenshots or code snippets**
N/A

**Desktop (please complete the following information):**
 - OS: Mac OSX
 - Version:  10.15.7

**What database are you using?**
 - Type: Postgres
 - Version 10.5
 - Masonite ORM: latest 

**Additional context**
N/A

girardinsamuel:
To solve this you should normally redefine the `DB_CONFIG_PATH` environment variable in your `.env` file to:
```bash
DB_CONFIG_PATH=config/core/database
```

josephmancuso:
Yea this is the solution 👆 

@girardinsamuel is this not documented? I don't see it in the install part of Masonite orm

circulon:
> To solve this you should normally redefine the `DB_CONFIG_PATH` environment variable in your `.env` file to:
> 
> ```shell
> DB_CONFIG_PATH=config/core/database
> ```

This is fine if your using .env files which I am not.

To clarify I am building a webapp in the cloud (AWS Lambda)
All environment variable on a given Lambda are visible when the lambda is viewed in the console.

This is obviously not desirable for sensite items like passwords and external api keys (eg Stripe)

In my case the config files all pull their item info from AWS Parameter Store (SSM) and so is only available to the code at runtime.

my desired config folder structure looks like this:

config
  - core
  - - application.py
  - - database.py
  - - cache.py
  - - ...
  - environment
  - - dev
  - - - application.py
  - - - database.py
  - - - cache.py
  - - - ...
  - - staging
  - - - application.py
  - - - ...
  - - prod
  - - - ...
  - - deploy 
  - - - ....

The config gets loaded in the following order and each layer has the ability to override the setting if set previously.
- config/core
[...]

PR #642, opened 2022-04-23
Handle using a custom path for the config file

circulon:
This PR covers using a custom path for the database config.

Incomplete fix for #641 

At this stage I could not easily work out how to pass the "config_path" to the load_config() in the Base Model class.

Happy to discuss / adjust  as required


josephmancuso:
Pushed a fix for broken tests

josephmancuso (approved):
@circulon Code looks good. Just to be clear, you plan to pass the config path onto the connection resolver? This seems backwards compatible to how it works now but i'm not sure how you plan to use this to work for your custom config paths?

josephmancuso:
Any update on this? I'm not sure how this would be used? @circulon

josephmancuso:
Just bumping this again. I don't see how passing a config path in the connection resolver will help solve the path issue. The class goes in the file its trying to find. So its a catch 22 situation i think

josephmancuso:
Closing this. Tried to figure out how to use this and wasn't able to. Setting the config path in the connection resolver won't work i think because its the file that this class goes in that we are looking for to begin with

PR #643, opened 2022-04-24
Added .first_or_create() method to Model.

JarriqTheTechie:
Retrieves first record matching a given criteria. If no record exists matching the given criteria, a new record is created.


JarriqTheTechie:

josephmancuso (approved):

PR #644, opened 2022-04-24
Added .first_or() method to Model.

JarriqTheTechie:
Retrieves first record matching a given criteria. If no record exists matching the given criteria, a user-defined callback function will be executed.

Example usage. 

```
def send_mail():
    return "mail sent"

User.first_or({"name": "bob"}, lambda: send_mail())
```
If no user with the name bob is found then callback function `send_mail()` will be executed.


JarriqTheTechie:
@josephmancuso ready for review

PR #647, opened 2022-04-25
Added .first_or() method to Model

JarriqTheTechie:
Retrieves first record matching a given criteria. If no record exists matching the given criteria, a user-defined callback function will be executed.

Example usage.

```
def send_mail():
    return "mail sent"
```

`User.first_or({"name": "bob"}, lambda: send_mail())`
If no user with the name bob is found then callback function `send_mail()` will be executed.

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -564,6 +564,21 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
         return json.dumps(self.serialize())
 
     @classmethod
+    def first_or(cls, wheres, callback=None, *args):
+        """Execute the query and get the first result or call a callback.
+
+        Returns:
+            Record|Mixed
+        """
+        self = cls()
+        self._check_is_callable(callback)
+        record = self.where(wheres).first()
+        if not record:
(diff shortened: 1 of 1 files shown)


JarriqTheTechie:

src/masoniteorm/models/Model.py
```diff
@@ -563,6 +563,21 @@ def to_json(self):
...
 
+    @classmethod
+    def first_or(cls, wheres, callback=None, *args):
+        """Execute the query and get the first result or call a callback.
+
+        Returns:
+            Record|Mixed
+        """
+        self = cls()
+        self._check_is_callable(callback)
```
josephmancuso, line 606: what is this doing? You built a check for is callable but not using the return value from it anywhere

josephmancuso (changes requested):

src/masoniteorm/models/Model.py
josephmancuso: @JarriqTheTechie any update on this? This is wrong right?

josephmancuso:
@JarriqTheTechie any update on these comments? I'd like to get this PR merged

PR #649, opened 2022-05-02
Fix/637 - Fixed updating pivot table records on Many To Many relationships

federicoparroni:


josephmancuso (changes requested):
Even though you couldn't replicate it. Should do the same thing to the `apply_query` method. I think that gets called when you eager load?

federicoparroni:
Yes, I can add it to the `apply_query` method too. In the meanwhile, could you please add some tests to cover also this cases?

federicoparroni:
@josephmancuso fixed also the `apply_query` method

federicoparroni:
@josephmancuso hello! Do you think that this PR can be merged?

josephmancuso (approved):

josephmancuso:
@federicoparroni yes it looks good. I couldn't get to this yesterday and just want to test this out before merging it. But the code review looks good 👍

federicoparroni:
@josephmancuso thank you so much for everything you did!

PR #650, opened 2022-05-05
Fixed string type casting/coercion from `where_in` methods

stoutput:
Masonite should not cast args to strings when compiling SQL, or when checking if some data exists in a collection. Rather it should use the raw data that was passed (`where()` already does this), and not make assumptions about what kind of data it is. Special cases should be handled manually, such as numeric strings matching integers.

This allows for using data such as bytes objects as args to `where_in` clauses, which will now correctly match against all database-supported data types.

This _may_ break some callers (that expect string coercion), so it should probably be considered backward-incompatible.

### DIFF
--- a/src/masoniteorm/collection/Collection.py
+++ b/src/masoniteorm/collection/Collection.py
@@ -409,10 +409,9 @@ class Collection:
                 attributes.append(item)
         return self.__class__(attributes)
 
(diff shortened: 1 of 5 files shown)


josephmancuso (approved):
LGTM and all tests are passing 👍

PR #651, opened 2022-05-05
Fixed edge-case Collection `where_in` bugs

stoutput:
Fixes a couple small edge-case bugs that would incorrectly match `False` & `None` values in collections when using `where_in()`


josephmancuso (approved):
LGTM 👍

PR #662, opened 2022-05-11
Fix local tests

stoutput:
This is a very basic docker-based solution for getting tests running locally


stoutput:
There's obviously still an issue with `MYSQL_ALLOW_EMPTY_PASSWORD`, and there should probably be a delay or wait after starting the containers but before running tests in `make test`

josephmancuso:
I made some changes but looks like an issue with the GitHub action still before I merge this in

PR #698, opened 2022-05-30
Added support to having_raw builder method

felipehertzer:
This method will allow to use subqueries, functions and select variables inside the having expression.

```python
builder.select_raw("COUNT(*) as counts").having_raw("counts > 10").to_sql()
```


josephmancuso (approved):
Awesome. LGTM 👍

PR #706, opened 2022-06-02
Added `model:docstring` command 

### LINKED ISSUE #240: Add list_table_columns to schema builder

### LINKED ISSUE #705: Add a "model:docstring" command to generate type hinting for models
To have type hinting for model attributes what we can do is to add a new command `model:docstring` to generate automatically type hints for the models. Then we would just have to copy/paste generated hints in our model classes.

This command existed before so we just have to write it back here. IMO it should live in the ORM codebase so that non Masonite users can benefit of it.

_Originally posted by @girardinsamuel in https://github.com/MasoniteFramework/orm/issues/695#issuecomment-1144566820_

girardinsamuel:
Fix #705
Closes #240
Based on PR https://github.com/MasoniteFramework/orm/pull/326.
TASK-55

Work in Progress...some types to correct and also fetching default values. Also need to add mysql and mssql support.


josephmancuso (approved):

josephmancuso:
Not doing MSSQL for now is fine.

josephmancuso (approved):

PR #709, opened 2022-06-05
fixed add binding issue

### LINKED ISSUE #702: Where_has function can only works at boolean but number nor string
**Describe the bug**
When using where_has function query, it only works at boolean condition but number nor string.

**To Reproduce**
1. Set Model:
   <pre><code>
    Class BillingAccountProjectContracts(Model):
    @has_one('id', 'contract_id')
    def contract(self):
         return models.Contract
    @scope
    def ifSupport(self, query):
        return query.where_has(
            "contract", lambda query: query.where("support_enabled", True)
        ).get()
    @scope
    def active(self, query):
        return query.where_has(
            "contract", lambda query: query.where('status', 1)
        ).get()
    @scope
    def ifTWD(self, query):
        return query.where_has(
            "contract", lambda query: query.where('currency', 'twd')
        ).get()
    </code></pre>
3. call defined function:
     obj = models.BillingAccountProjectContracts.active().get() 
     obj = models.BillingAccountProjectContracts.ifTWD().get() 

**Expected behavior**
Return collection which match the conditions.

image

 - OS: Mac OS Monterey
 - Version 12.3 with Apple M1

 - Type: Postgres
 - Version 12.10
 - Masonite ORM v2.11.9

**Additional context**
Error won't happen and get the right results when calling models.BillingAccountProjectContracts.ifSupport().get() function.

josephmancuso:
Closes #702

### DIFF
--- a/src/masoniteorm/query/grammars/BaseGrammar.py
+++ b/src/masoniteorm/query/grammars/BaseGrammar.py
@@ -647,7 +647,7 @@ class BaseGrammar:
                 if qmark:
                     query_from_builder = value.builder.to_qmark()
                     if value.builder._bindings:
-                        self.add_binding(value.builder._bindings)
+                        self.add_binding(*value.builder._bindings)
                 else:
                     query_from_builder = value.builder.to_sql()
                 query_value = self.subquery_string().format(query=query_from_builder)


josephmancuso (commented):
reviewed

PR #721, opened 2022-06-12
Added new method | get_column_listing()

JarriqTheTechie:
Get the column listing for a given table.

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -157,6 +157,7 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
         "from_",
         "from_raw",
         "get",
+        "get_column_listing",
         "group_by_raw",
         "group_by",
         "has",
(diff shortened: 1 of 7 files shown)


JarriqTheTechie:
@josephmancuso Ready for review!

josephmancuso (commented):
I think this functionality is being done in https://github.com/MasoniteFramework/orm/pull/706 so i don't want to duplicate work

josephmancuso:
although slightly different in https://github.com/MasoniteFramework/orm/pull/706 (this is just getting column names when #706 is getting the full table definition for another open issue) the functionality is the same and you should be able to just pluck out the column keys from the schema built

JarriqTheTechie:
Okay sounds good. Will #706 work in QueryBuilder and Model? That's the primary reason for the PR. I'm planning to roll out my own adminpanel and that functionality would be awesome if built-in.

josephmancuso:
right that's what this does. It wouldn't in terms of the requirement of that PR but it would be able to. 

Right now its on the schema class so now you would do this

```python
from config.database import DB

columns = DB.get_schema_builder().get_columns("users")
```

The query builder should have access to that DB class so we can build a query builder method around this

josephmancuso:
I'm testing #706 now. Its a draft but i think its done and if its not i'll wrap things up on it

JarriqTheTechie:
Exquisite. I'll roll with with this implementation until #706 is merged. Great work.

PR #722, opened 2022-06-12
Added new method has_column()

JarriqTheTechie:
Added new method has_column()


josephmancuso:
Ok you can either close this PR and open a new one or change the code to match the new way.

Should be something like this on the QueryBuilder class now:

```python
def has_column(self, name):
    return self. get_table_columns().get(name)
```

josephmancuso:
is this PR a duplicate of #721

PR #723, opened 2022-06-12
Feature/has columns

JarriqTheTechie:
Added new feature has_columns()

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -161,6 +161,7 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
         "group_by_raw",
         "group_by",
         "has",
+        "has_column",
         "having",
         "having_raw",
         "increment",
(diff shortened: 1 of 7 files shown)


Marlysson:
@JarriqTheTechie 

The Schema class already has the "has_column" method implemented, this work for you case?

https://github.com/MasoniteFramework/orm/blob/2.0/src/masoniteorm/schema/Schema.py#L170-L185

josephmancuso:
The schema class does but it only works for Sqlite right now.

706 builds out the rest of the databases

josephmancuso:
Sqlite migrations are extremely complex. There's basically no alter commands in SQLite. so to make any table changes we actually have to create temporary tables, drop the existing table, create a new table and migrate the temp table data into the new table. So we had to build out the schema information into a structure that would support doing that.

#706 builds out this structure for MySQL & Postgres (while also satisfying the requirement for the model docstrings that M3 had)

JarriqTheTechie:
I'm so used to doing everything from QueryBuilder & Model that if I don't see it there I automatically assume it doesn't exist.

josephmancuso:
> I'm so used to doing everything from QueryBuilder & Model that if I don't see it there I automatically assume it doesn't exist.

I built everything in this library on a modular way so if the theres nothing on the query builder than its just not attached to it yet

josephmancuso:
Will need to reopen a new PR without all the additions to the grammars

PR #724, opened 2022-06-12
fixed how scopes work

### LINKED ISSUE #720: Issue when serializing a model with pickle
```python
from databases.models import Account
import pickle 

accounts = Account.is_logged_in().in_random_order()\
    .take(100).get().chunk(10)

pickle.dumps(accounts)
```

Error:
```text
_pickle.PicklingError: Can't pickle <function Account.is_logged_in at 0x7f6b36016a70>: it's not the same object as databases.models.Account.Account.is_logged_in
```

is_logged_in is a model scope

josephmancuso:
Closes #720

### DIFF
--- a/src/masoniteorm/query/QueryBuilder.py
+++ b/src/masoniteorm/query/QueryBuilder.py
@@ -35,7 +35,6 @@ import pendulum
 from ..schema import Schema
 
 
-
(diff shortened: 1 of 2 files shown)


josephmancuso:
No way to really make this work right now

PR #725, opened 2022-06-12
Added a distinct method to query builder for specifying distinct SELECT queries

### LINKED ISSUE #718: Add select distinct
Need to add a select distinct.

should probably be something like this:

```python
User.select("name").distinct().get()
```

josephmancuso:
This PR adds the behavior to add a distinct select clause:

```python
User.select("email").distinct().get()
```

Closes #718


josephmancuso (commented):
approved

PR #729, opened 2022-06-18
Always check column casts when creating

circulon:
fixes issue 729


josephmancuso:
we'll have to pluck this change out when we start development on version 3

PR #735, opened 2022-06-25
Feature/684 - Changed using raw expressions to use the ? placeholder rather than a '?' placeholder

### LINKED ISSUE #684: Masonite ORM should change '?' to ? in the binding expressions
One of the smaller odd things with Masonite ORM was the binding syntax. It needed a `'?'` in the query statements when trying to bind raw statements.

Example:

```python
builder.statement("select count(*) from users where active = '?'", [1])
```

Notice the quotes around the `?` mark. It would be cleaner to be able to just use a `?` mark.

Would also need to be done in the grammars as well. And in the connection classes we are replacing this with that the query connection needs

This change would have to be done with Masonite ORM 3 since it would be a breaking change. Can do this issue as we gear up to Masonite 5

https://orm.masoniteproject.com/query-builder#raw-queries

josephmancuso:
### Changed

* Changed using raw expressions to use the `?` placeholder rather than a `'?'` placeholder

Closes #684

### DIFF
--- a/src/masoniteorm/connections/MSSQLConnection.py
+++ b/src/masoniteorm/connections/MSSQLConnection.py
@@ -145,7 +145,6 @@ class MSSQLConnection(BaseConnection):
                     for q in query:
                         self.statement(q, ())
                     return
-                query = query.replace("'?'", "?")
                 self.statement(query, bindings)
                 if results == 1:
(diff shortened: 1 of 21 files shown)


josephmancuso (commented):
reviewed

PR #738, opened 2022-06-27
Feature/682

### LINKED ISSUE #682: QueryBuilder created with `new_from_builder()` use the same references as the builder it is created from
Hi,
we are trying to do something like this:

```python
website_query = Website.with_trashed()

# apply filter
website_query = apply_filter(website_query, filter)

# total items
total_items = website_query.new_from_builder().count()

# pagination
if limit and offset is not None:
    website_query = website_query.limit(limit).offset(offset)

# query execution
websites = website_query.get()
```

where `apply_filter()` besides some wheres, also adds a subquery to the builder:

```python
operator, value = query._extract_operator_value(fil["op"], fil["value"])
subquery = type(query._model)
builder = getattr(subquery, f'filter_{fil["field"]}')(fil["op"], fil["value"] )
query._wheres += (
        (QueryExpression(None, operator, SubGroupExpression(builder), keyword=kw)),
)
```

The error that we have is that after doing `count()` (even though we use `new_from_builder`) the query goes from this:
```sql
SELECT * FROM `apps` WHERE ( EXISTS (SELECT * FROM `users` WHERE `users`.`id` = `apps`.`user_id` AND `users`.`username` = '<username>' AND `users`.`id` > '99')) AND `apps`.`archived_at` IS NULL
```
to this:
```sql
SELECT * FROM `apps` WHERE ( EXISTS (SELECT * FROM `users` WHERE `users`.`id` > '99')) AND `apps`.`archived_at` IS NULL
```

We think the problem is that `new_from_builder()` is creating a new query builder, but adding the data from the old one with reference, so when `count` executes `get()` (and get resets the query builder in the end), it also resets the _wheres part of the old query builder. We think it should somehow do a deep copy, but whatever we tried, we didn't manage to solve it.

Desktop:

    OS: Linux
    Version Ubuntu 21.10

What database are you using?

    Type: [MySQL]
    Version [14.14 Distrib 5.7.30, for Linux (x86_64) using EditLine wrapper]
    Masonite ORM [2.4.2]

JRubics:
Fix new_from_builder using deepcopy but with fixing the infinite recursion

Fixes #682

### DIFF
--- a/src/masoniteorm/query/QueryBuilder.py
+++ b/src/masoniteorm/query/QueryBuilder.py
@@ -356,7 +356,9 @@ class QueryBuilder(ObservesEvents):
             self
         """
         if attribute == "__setstate__":
-            raise AttributeError(attribute)
+            raise AttributeError(
+                "'QueryBuilder' object has no attribute '{}'".format(attribute)
+            )
 
         if attribute in self._scopes:
 
--- a/src/masoniteorm/query/QueryBuilder.py
+++ b/src/masoniteorm/query/QueryBuilder.py
@@ -1,5 +1,5 @@
 import inspect
-from copy import deepcopy, copy
+from copy import deepcopy
 
 from ..config import load_config
 from ..collection.Collection import Collection
@@ -355,6 +355,9 @@ class QueryBuilder(ObservesEvents):
         Returns:
             self
         """
+        if attribute == "__setstate__":
+            raise AttributeError(attribute)
+
         if attribute in self._scopes:
 
             def method(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


josephmancuso:
ugh thanks for fixing this lol. Interesting fix I didn't know `__setstate__` was a magic method like this. I have another issue where an object wasn't pickleable. Maybe using these magic methods will fix that issue too 🤔

JRubics:
I think it is at least worth trying and as I remember there was some magic method for pickling, but don't remember exactly how to use it :slightly_smiling_face:

PR #745, opened 2022-07-03
added schema for postgres

### LINKED ISSUE #737: Add Schema to postgres connections
Should be able to specify a schema at the connection level (which would then be able to be modified on the fly)

We actually cannot specify the schema at all. It will always default to the `default` schema

**Additional context**
In the future this will need to be changed via a middleware so schemas can change based on he hostname or something

- [ ] Is this a breaking change? - no should not be

josephmancuso:
Closes #737


josephmancuso (commented):
reviewed

PR #754, opened 2022-07-08
Fixed JSON Cast should always return a string

circulon:
This fixes the issue where a valid JSON string was not returned if the value being cast was a dict or list.
Added validation (as a valid JSON structure) of the value if it is a string


josephmancuso (approved):

if4lcon:
@josephmancuso this PR made a bug ... when you get attribute you want it to be json loaded but idk why json.loads was removed

PR #757, opened 2022-07-11
Fix/where has

### LINKED ISSUE #756: with_ method should be able to use a lambda
Should be able to do syntax like this:

```python
User.with_({"phone": lambda q: q.limit(10)}).get()
```

josephmancuso:
Closes #756


josephmancuso (commented):
reviewed

josephmancuso (commented):
reviewed

PR #774, opened 2022-07-27
Fixed wrong select structure when using having and order by

felipehertzer:
Hi,

This pull request is to fix a problem in the structure of select when using having and order.

ORDER, HAVING to HAVING, ORDER

```python
QueryBuilder().table("user").where_null("deletedat").group_by("age").having_raw("age > 2").order_by("age")
```
Current:
```sql
SELECT * FROM `user` WHERE `user`.`deletedat` IS NULL GROUP BY `user`.`age` ORDER BY `age` ASC HAVING age > 2
```
Should be:
```sql
SELECT * FROM `user` WHERE `user`.`deletedat` IS NULL GROUP BY `user`.`age` HAVING age > 2 ORDER BY `age` ASC
```

Source: https://www.w3schools.com/sql/sql_having.asp

### DIFF
--- a/src/masoniteorm/query/grammars/MSSQLGrammar.py
+++ b/src/masoniteorm/query/grammars/MSSQLGrammar.py
@@ -10,7 +10,6 @@ class MSSQLGrammar(BaseGrammar):
         "MIN": "MIN",
         "AVG": "AVG",
         "COUNT": "COUNT",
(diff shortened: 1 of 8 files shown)


josephmancuso (approved):
Looks good to me

josephmancuso (approved):

PR #779, opened 2022-07-31
changed how collection is plucked in relatinoships

### LINKED ISSUE #777: Column 'id' cannot be null issue during observers
When using eager loading on the documentation it's stated that the number of queries on DB will be just one, but using eager loading when having a relationship not hidden (i.e. serializable ) the number of queries with eager loading is greater than without eager loading.

**Example To Reproduce**
I created a small repo to illustrate the problem 
https://github.com/lucadelmonte/masonite_eager_bug
you can run the run.sh file, it will create the tables on a sqllite db.

**Expected behavior**
Eager loading should only execute 2 queries, one for the Parent model and one for the childs (one for relationship)

 - OS:Linux
 - Version: Ubuntu 21.10
 
tested on both mysql, mariadb and SqlLite
Masonite ORM [2.15.0]

**Possible solution**
The additional queries are executed by https://github.com/MasoniteFramework/orm/blob/da41d341fc2604346261c695651ac71a346408ec/src/masoniteorm/relationships/BaseRelationship.py#L155 the call to pluck on BaseRelationship which in turn calls serialize on all the child objects before executing the "eager loading query"` select * from X IN (....)`, i managed to fix the problem replacing the call to pluck with a one to `Collection(relation._get_value(self.local_key)).unique()`https://github.com/MasoniteFramework/orm/blob/da41d341fc2604346261c695651ac71a346408ec/src/masoniteorm/collection/Collection.py#L457

**Additional**
Can you please apply the patch to the version 2.15.0, because we tried to upgrade to 2.18.2 and it has some incompatibilities.

josephmancuso:
Closes #777

### DIFF
--- a/src/masoniteorm/relationships/BaseRelationship.py
+++ b/src/masoniteorm/relationships/BaseRelationship.py
@@ -152,7 +152,7 @@ class BaseRelationship:
         if isinstance(relation, Collection):
             return builder.where_in(
                 f"{builder.get_table_name()}.{self.foreign_key}",
-                relation.pluck(self.local_key, keep_nulls=False).unique(),
+                Collection(relation._get_value(self.local_key)).unique(),
             ).get()
         else:
             return builder.where(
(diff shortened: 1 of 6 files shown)


josephmancuso (commented):
reviewed

PR #786, opened 2022-08-30
Fix record creation in model observer

federicoparroni:


federicoparroni:
That issue can be solved by inverting the attach logic:
```python
class Restore(Model):
    # ....
    def createFallback(self):
        backup = Backup.create({
            'website_id': self.backup.website_id,
            'name': 'Automatic backup before restore',
        })
        # self.attach('fallback', backup)  # changed this
        backup.attach('restore', self)     # to this
        self.save()
```

PR #795, opened 2022-09-21
Adds oldest, latest methods to QueryBuilder, fixes #790

### LINKED ISSUE #790: Add some methods
Add latest and oldest methods to query builder

yubarajshrestha:
fixes MasoniteFramework/orm#790

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -255,6 +255,8 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
         "where_doesnt_have",
         "with_",
         "with_count",
+        "latest",
+        "oldest"
     ]
 
     __cast_map__ = {}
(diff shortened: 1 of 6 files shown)


josephmancuso (changes requested):
I don't really like the order by raw solution here. If we do the order by raw it won't work with postgres or sql server

```python
for field in fields:
    self.order_by(field, "desc")
```

yubarajshrestha:
> I don't really like the order by raw solution here. If we do the order by raw it won't work with postgres or sql server
> 
> ```python
> for field in fields:
>     self.order_by(field, "desc")
> ```

Updated to `order_by`.

BSN4:
tests?

josephmancuso:
This looks better but will need some tests to make me feel comfortable with it

josephmancuso (changes requested):
This looks better but will need some tests to make me feel comfortable with it

yubarajshrestha:
> This looks better but will need some tests to make me feel comfortable with it

Done

yubarajshrestha:
> This looks better but will need some tests to make me feel comfortable with it

Done

josephmancuso (approved):

PR #809, opened 2022-10-13
added delete_quietly method on Model

JarriqTheTechie:
This method calls the delete method on a model without firing the delete & deleting observer events.
Instead of calling:

User().delete(...)

you can use this:

User.delete_quietly(...)

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -1066,7 +1066,9 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
         Returns:
             self
         """
-        return self.without_events().delete()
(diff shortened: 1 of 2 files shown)


JarriqTheTechie:
@girardinsamuel / @josephmancuso please review.

Marlysson:
Shouldn't activate again the events after this operation as save and updte quietly does with `.with_events()` method?

josephmancuso (changes requested):
@Marlysson is right here. You need to set `with_events()` again.

josephmancuso (approved):
Changed a few things. Your change would have deleted the whole table

JarriqTheTechie:
Thanks @Marlysson @josephmancuso

PR #810, opened 2022-10-13
added save_quietly method on Model

JarriqTheTechie:
This method calls the save method on a model without firing the saved & saving observer events. Saved/Saving are toggled back on once save_quietly has been ran.

Instead of calling:

User().save(...)

you can use this:

User.save_quietly(...)


JarriqTheTechie:
@josephmancuso / @girardinsamuel please review.

josephmancuso (approved):
Fixed same issue with this PR

josephmancuso (approved):

PR #811, opened 2022-10-13
added update_quietly method on Model

JarriqTheTechie:
This method calls the update method on a model without firing the updated & updating observer events. Updated/Updating are toggled back on once update_quietly has been ran.

Instead of calling:

User().update(...)

you can use this:

User.update_quietly(...)

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -1053,6 +1053,23 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
 
         return related.detach(self, related_record)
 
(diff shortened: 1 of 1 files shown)


JarriqTheTechie:
@josephmancuso / @girardinsamuel please review.

josephmancuso (changes requested):
This should be `update_quietly` and do the same changes i did in the other PR's

JarriqTheTechie:
Hi @josephmancuso, I'll make the changes this weekend. Been having a boxing match with COVID.

josephmancuso:
of course no worries just doing some cleaning up

PR #812, opened 2022-10-13
Fix turn on flag in with_events method

Marlysson:


josephmancuso (commented):
Thanks @Marlysson. I actually got this in another PR i was fixing so this can be closed

PR #818, opened 2022-10-20
JSONCast Force conversion to str for unhandled types

circulon:
This fixes issues when casting types like Decimal to JSON throwing a TypeError

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -72,7 +72,7 @@ class JsonCast:
             json.loads(value)
             return value
 
-        return json.dumps(value)
+        return json.dumps(value, default=str)
 
 
 class IntCast:


BSN4:
better to convert it to float

```python
class JSONEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, Decimal):
            return float(obj)
        return json.JSONEncoder.default(self, obj)
```

circulon:
> better to convert it to float
> 
> ```python
> class JSONEncoder(json.JSONEncoder):
>     def default(self, obj):
>         if isinstance(obj, Decimal):
>             return float(obj)
>         return json.JSONEncoder.default(self, obj)
> ```

@if4lcon 
No you miss the point, Decimal is just an example and is not the only use case here.
Other examples  are UUID
so by defining the `default` any obect that is not directly JSON encodable will be asked for its `str` repr which is encodable.

Also converting a Decimal to float loses specificity of the value so that would not be ideal, but that is another conversation

I hope this clarifies things for you

josephmancuso (approved):
Looks fine to me

PR #823, opened 2022-10-26
Adding stubs and html

NawrasBukhari:
I had been using this ORM for a project when I discovered that when using it as a standalone it does not automatically include the stubs Due to this issue, I got errors in that file.stub was not found. for reference when using it as a standalone I leave the following repo https://github.com/NawrasBukhari/pyt-graduation-project


josephmancuso (approved):

josephmancuso:
merge is fsiling you might need to pull and reopen

PR #825, opened 2022-10-27
fix: Belongs to many - Table can't be without an underscore

### LINKED ISSUE #803: Belongs to many - Table can't be without an underscore
**Describe the bug**
In a belongs to many relationship, the table option can't be without an underscore (i.e. `compositions`), but it should be `table1_table2`. If this structure isn't present it throws an error when splitting the table name. This behaviour isn't necessary when `local_foreign_key` and `other_foreign_key` are set.

**To Reproduce**
1. Create a sample database with two tables and one "belongstomany" table
2. Create the models for the table
3. Use the code below for the "belongs to many" relationship
4. Try to get the related model & see error

**Expected behavior**
Table works with the two additional options

```python
@belongs_to_many(local_foreign_key="donut_id", other_foreign_key="", pivot_id=None, table="compositions", with_fields=["absolute_quantity"])
    def ingredients(self):
        """Get the ingredients for this donut"""
        from app.Models.Ingredient import Ingredient
        return Ingredient
```

 - OS: Windows
 - Version 11

 - Type: MariaDB
 - Version 10.6.9
 - Masonite ORM 2.18.6

maicol07:
Fixes #803

### DIFF
--- a/src/masoniteorm/relationships/BelongsToMany.py
+++ b/src/masoniteorm/relationships/BelongsToMany.py
@@ -67,7 +67,7 @@ class BelongsToMany(BaseRelationship):
             self._table = "_".join(pivot_tables)
             self.foreign_key = self.foreign_key or f"{pivot_table_1}_id"
             self.local_key = self.local_key or f"{pivot_table_2}_id"
-        else:
+        elif self.local_key is None or self.foreign_key is None:
             pivot_table_1, pivot_table_2 = self._table.split("_", 1)
             self.foreign_key = self.foreign_key or f"{pivot_table_1}_id"
             self.local_key = self.local_key or f"{pivot_table_2}_id"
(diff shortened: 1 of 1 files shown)


josephmancuso (changes requested):
This won't work in the scenario that someone only specifies the first key right? like if someone specifies the self.local_key but not the self.foreign_key.

In that case i think it would overwrite whatever they specified as the self.local_key

maicol07:
> This won't work in the scenario that someone only specifies the first key right? like if someone specifies the self.local_key but not the self.foreign_key.
> 
> In that case i think it would overwrite whatever they specified as the self.local_key

Why? Inside the `elif` there's the `or` that should keep the key id it's defined

maicol07:
@josephmancuso ↑

josephmancuso (approved):

PR #828, opened 2022-10-28
Adding RETURNING * clause to return all fields after creation model.

### LINKED ISSUE #827: Returning all fields after creation even if it's NULL
**Describe the bug**
Is there a way to hydrate model that includes null fields? Because if I load the model from DB, all fields that are null are not included in Model.__attributes__ 
Even Model.all_attributes() doesn't give me all attributes hydrated with None values.

**To Reproduce**
```python
result = Model.create(field=value)
result.__attributes__ # nullable field with NULL values will not appear in dict
```
**Expected behavior**
```python
result.__attributes__ -> {...., {"nullable_field": None}}
```

Marlysson:
Closes #827 
To return all fields after creation model.
@girardinsamuel @josephmancuso

### DIFF
--- a/src/masoniteorm/observers/ObservesEvents.py
+++ b/src/masoniteorm/observers/ObservesEvents.py
@@ -23,5 +23,5 @@ class ObservesEvents:
     @classmethod
     def with_events(cls):
         """Sets __has_events__ attribute on model to True."""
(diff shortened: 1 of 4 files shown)


josephmancuso:
Needs to fix tests

PR #830, opened 2022-11-17
Fix fillable and guarded attribute behavior on mass-assignment

stoutput:
Previously, the `__fillable__` and `__guarded__` attributes were only partially supported, and only on create.

This PR:
- adds full support for the `__fillable__` and `__guarded__` attributes on `create`, `bulk_create`, and `update`
- corrects the default value of `__guarded__` to be `[]`
    - `[]` means no attributes are guarded against mass assignment, thus allowing any field to be set by `update`/`create`
    - `['*']`, by contrast, means that all fields on the model are guarded against mass assignment and cannot be set with `update`/`create`
    - It's possible that you may not want fields filtered on `QueryBuilder.create()` since this is a more "direct" method call and would still allow for creation if all the fields are excluded by fillable/guarded. If so let me know and I can remove this behavior
- now raises if both `__fillable__` and `__guarded__` are defined on the base model. This mirrors Orator's behavior.
- fleshes out full support of optional casting of values during `create`/`bulk_create`/`update`
- cleans up `requirements.txt` to only install runtime dependencies. Dev dependencies have been moved into `requirements.dev` and the makefile adjusted
- updates `black` due to a bug with one of its dependencies, `click` (in order to run formatting for this PR)
- cleans up a whole lot of code along the way

Note: I still can't run tests locally in a fresh environment, which makes contributing well-tested code somewhat difficult. Was getting the error about a missing ORM config file.

Update: I was able to get most tests passing locally by creating a database config file. Its path gets set during testing only.


stoutput:
@josephmancuso This is ready for review. Sorry, I know it's a big one.

BSN4:
looks nice, so many code got refactored though need to be reviewed and tested for sure

stoutput:
@josephmancuso Think you can review this one soon? Thanks!

josephmancuso:
Yes my apologies ill get to this. Its on my TODO

zombeer:
I'm also waiting for this fix

src/masoniteorm/config.py
```diff
@@ -11,11 +11,16 @@ def load_config(config_path=None):
         1. try to load from DB_CONFIG_PATH environment variable
         2. else try to load from default config_path: config/database
     """
+    env_path = os.getenv("DB_CONFIG_PATH", None)
 
-    if not os.getenv("DB_CONFIG_PATH", None):
-        os.environ["DB_CONFIG_PATH"] = config_path or "config/database"
+    if env_path:
+        selected_config_path = env_path
+    elif os.getenv("TESTING", "").lower() == "true":
```
josephmancuso: I think we need to name this something a little more obscure. This is only for developing the library right? If someone in their application has a environment variable calling TESTING (very possible) it will break if they don't have a config/test-database.py file. Can we name this something like `ORM_TESTING`?

josephmancuso (changes requested):

stoutput:
Ready for a second pass @josephmancuso

.envrc
```diff
@@ -0,0 +1,2 @@
+use asdf
+layout python
```
josephmancuso, line 2: Whats this file for?

.github/workflows/pythonapp.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           python orm migrate --connection mysql
           make test
   lint:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     name: Lint
```
josephmancuso, line 62: Whats the point of pegging this to a version instead of always testing against the latest ubuntu version?

josephmancuso (changes requested):
Just want to understand the CI changes you made more

.github/workflows/pythonapp.yml
josephmancuso: If we always use the latest we can catch issues as soon as the latest version is updated on that container. If we always peg to 20.04 we'll know it works on ubuntu 20.04 but never anything newer than that

stoutput:
@josephmancuso replied to your comments – let me know if you have any other questions!

stoutput:
Hey @josephmancuso I'm sure you're super busy but it'd be great if we didn't let this one languish, there are a few people waiting on this fix. If you want, I'd be open to taking on a co-maintainer role of the ORM repo while you're busy.

PR #835, opened 2023-01-31
Migrations should ignore dotfiles.

bezborodow:
Bugfix for **https://github.com/MasoniteFramework/masonite/issues/743**.


bezborodow:
This will ignore all filenames that begin with `.`.

For completeness, it might also be best to only run migration filenames that end in `.py`.

josephmancuso (approved):

PR #837, opened 2023-02-01
Feature/find_or

JarriqTheTechie:
Retrieves record by primary key. If no record exists matching the given criteria, a user-defined callback function will be executed.

eg.

id: int = 1024
callback: Callable = lambda a, b : a * b
arguments: tuple = (5, 6)

User.find_or(id, callback, arguments)

The callback will run only if user with id 1024 is not found. Arguments to the callback are passed in as tuples.

### DIFF
--- a/src/masoniteorm/exceptions.py
+++ b/src/masoniteorm/exceptions.py
@@ -32,3 +32,7 @@ class InvalidUrlConfiguration(Exception):
 
 class MultipleRecordsFound(Exception):
     pass
+
+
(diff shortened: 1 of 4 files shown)


JarriqTheTechie:
@josephmancuso please review.

josephmancuso (changes requested):
needs a test

JarriqTheTechie:
Test has now been added. @josephmancuso please review.

josephmancuso (approved):
code review looks good

PR #839, opened 2023-02-15
Use dynamic model variables for timestamp column names

stoutput:
Replaces hardcoded `"created_at"` and `"updated_at"` timestamp column names with base model's `date_created_at` and `date_updated_at` variables. This allows configurability over which columns to use with the timestamps mixin. Besides, if we're already using these variables for date casting, we should also use them in the mixin.


stoutput:
@josephmancuso What do you think? Been waiting on this one for a while to finish out a PR in another repo.

stoutput:
I seem to recall dynamic timestamp column names had been implemented at one point in the past, but I'm not totally sure...

josephmancuso (approved):

PR #845, opened 2023-03-11
Fixed UUID PK Mixin not creating primary key on bulk create

circulon:
This PR fixes the issue where bulk create would not create a primary key attribute if the UUIDPrimaryKeyMixin was used on the Model


stoutput (approved):
Nice! Thanks for this. This LGTM. @josephmancuso?

josephmancuso (approved):

josephmancuso:
LGTM too

Issue #843, opened 2023-02-20
masoniteorm.query.QueryBuilder.QueryBuilder.create()


moluwole:
Hello guys, I have this blogpost https://testdriven.io/blog/masonite-orm-fastapi/. 

I tried upgrading the dependencies of the post so as to have the latest versions but when I try to save to the database, I get this error
```
"masoniteorm.query.QueryBuilder.QueryBuilder.create() got multiple values for keyword argument 'id_key'"
```

The whole project works fine with the `2.18.6` version but anything higher than that results in the error above

josephmancuso:
Can you put the whole code snippet here for me please? and also possibly any stack traces

moluwole:
### StackTrace

```sh
Traceback (most recent call last):
  File "/home/moluwole/Documents/code/writing/fastapi-masonite/.env/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 407, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/moluwole/Documents/code/writing/fastapi-masonite/.env/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/home/moluwole/Documents/code/writing/fastapi-masonite/.env/lib/python3.10/site-packages/fastapi/applications.py", line 270, in __call__
    await super().__call__(scope, receive, send)
  File "/home/moluwole/Documents/code/writing/fastapi-masonite/.env/lib/python3.10/site-packages/starlette/applications.py", line 124, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/moluwole/Documents/code/writing/fastapi-masonite/.env/lib/python3.10/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/home/moluwole/Documents/code/writing/fastapi-masonite/.env/lib/python3.10/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/home/moluwole/Documents/code/writing/fastapi-masonite/.env/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/home/moluwole/Documents/code/writing/fastapi-masonite/.env/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
[...]

PR #849, opened 2023-04-07
Fixed #848

### LINKED ISSUE #848: Little issue in first_or_create method
Code:
```python
user = User.first_or_create(
    {'id': sender_id}
)
```

Error:
`masoniteorm.query.QueryBuilder.QueryBuilder.create() got multiple values for keyword argument 'id_key'`

botify-framework:


tpow:
As I wrote in my comment in #843, I feel like the problem is "caused by the Model.save adding the id_key when it calls create. This means Model.create gets the id_key in the kwargs, but then adds it again. Arguably, save doesn't need to add it." Are there any use-cases where passing the id_key to create is valuable? Is this necessary in first_or_create?

lamson-01 (approved):

josephmancuso:
this was fixed in https://github.com/MasoniteFramework/orm/pull/856

PR #854, opened 2023-07-27
Added gevent pool driver option to MySQL connection

olivernadj:
Performance improvement for gevent applications.


src/masoniteorm/connections/ConnectionResolver.py
```diff
@@ -24,6 +24,12 @@ def __init__(self):
         self.register(MySQLConnection)
         self.register(MSSQLConnection)
 
+        try:
```
josephmancuso, line 27: why is this wrapped around an import exception?

josephmancuso (changes requested):

josephmancuso:
only for gevent applications? i would rather prefer this be its own connection class then that you can register for only those applications using the connection resolver.

like this : https://github.com/MasoniteFramework/orm/blob/2.0/src/masoniteorm/connections/ConnectionResolver.py#L21

PR #855, opened 2023-09-07
fixed class model has no attribute id when using update_or_create

sfinktah:
The following code is otherwise unworkable:

```py
ebay_transaction = EbayOrderTransaction.update_or_create({'transaction_id': d['transaction_id']}, d)
ebay_order.attach('transactions', ebay_transaction)
```

as the primary id key is not returned, and one cannot call `fresh` on the result.  it is possible [but undesirable] to work around this in such a way as:

```py
ebay_order = EbayOrder.where({'order_id': d['order_id']}).first()
if not ebay_order:
    ebay_order = EbayOrder.create(d).fresh()
else:
    ebay_order.save()
```

this PR also fixes an issue (that no doubt occurs in many other places too) were

```py
        return cls.builder.create(	        return cls.builder.create(
            dictionary, id_key=cls.__primary_key__, cast=cast, **kwargs	            dictionary, cast=cast, **kwargs
        )
```

will raise an Exception as the library is already passing `id_key` via `kwargs`

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -562,8 +562,10 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
                 dictionary, query=True, id_key=cls.__primary_key__, cast=cast, **kwargs
             ).to_sql()
 
+        # when 'id_key' is already present in kwargs, the previous version raised
+        kwargs['id_key'] = cls.__primary_key__
         return cls.builder.create(
-            dictionary, id_key=cls.__primary_key__, cast=cast, **kwargs
+            dictionary, cast=cast, **kwargs
         )
 
     @classmethod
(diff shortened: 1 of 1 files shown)


src/masoniteorm/models/Model.py
```diff
@@ -711,9 +713,19 @@ def update_or_create(cls, wheres, updates):
...
+            #
+            # Without this we are reduced to performing a DIY update_or_create, e.g.:
+            #     ebay_order = EbayOrder.where({'order_id': d['order_id']}).first()
+            #     if not ebay_order:
+            #         ebay_order = EbayOrder.create(d).fresh()
+            #     else:
+            #         ebay_order.save()
+            return self.create(total, id_key=cls.get_primary_key()).fresh()
+
+        rv = self.where(wheres).update(total)
```
JarriqTheTechie, line 723: This doesn't need to have a variable assigned to it since it isnt being used anywhere. 

self.where(wheres).update(total)

josephmancuso: agreed

josephmancuso (approved):
i think this looks right

src/masoniteorm/models/Model.py
sfinktah: Sorry, I'm presently mentally stuck in PHP, and none of it looks right.  Least of all my PR.

In Laravel, we do this:

```php
    // $attributes = wheres, $values = updates
    public function updateOrCreate(array $attributes, array $values = [])
    {
        $instance = $this->firstOrNew($attributes);

        $instance->fill($values)->save();

        return $instance;
    }
    
    public function firstOrNew(array $attributes)
    {
        if (is_null($instance = $this->where($attributes)->first())) {
            $instance = $this->related->newInstance($attributes);
        }

        return $instance;
    }
```
I'm not comfortable with [any version] of the code that updates more than 1 row... though at the time, I only intended to modify the return in cases where the record required creation.  

For full consistancy, I guess something like this... forgive me if the last two lines aren't valid, they're best-guess-pseudocode.

```py
    def update_or_create(cls, wheres, updates):
        self = cls()
        record = self.where(wheres).first()
        total = {}
        total.update(updates)
        total.update(wheres)
        if not record:
            return self.create(total, id_key=cls.get_primary_key()).fresh()

        record.update(total)
        return record
```


Hope I'm making some sense.

yubarajshrestha:
@sfinktah fix the linter issues.

josephmancuso (approved):

src/masoniteorm/models/Model.py
```diff
@@ -709,9 +709,19 @@ def update_or_create(cls, wheres, updates):
...
+        return self.where(wheres).first()
```
josephmancuso, line 724: @sfinktah i'm actually reverting this change. the wheres could not be unique enough i think for this to work. its possible we could do 

```
return self.where(total).first()
```

but it would have to be tested

PR #859, opened 2023-10-22
add batch number to status command

eaguad1337:
The `migrate:status` now show the Batch number. 

image

### DIFF
--- a/src/masoniteorm/migrations/Migration.py
+++ b/src/masoniteorm/migrations/Migration.py
@@ -118,8 +118,10 @@ class Migration:
 
         database_migrations = self.migration_model.all()
         for migration in all_migrations:
-            if migration in database_migrations.pluck("migration"):
-                ran.append(migration)
+            if migration := database_migrations.where("migration", migration).first():
+                ran.append(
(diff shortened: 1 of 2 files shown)


src/masoniteorm/migrations/Migration.py
```diff
@@ -118,8 +118,10 @@ def get_ran_migrations(self):
 
         database_migrations = self.migration_model.all()
         for migration in all_migrations:
-            if migration in database_migrations.pluck("migration"):
-                ran.append(migration)
+            if migration := database_migrations.where("migration", migration).first():
```
josephmancuso: oRM is currently supporting python 3.6+. this would break on 3.6 and 3.7 right?

josephmancuso (changes requested):

src/masoniteorm/migrations/Migration.py
eaguad1337: You're right, but this operator can be replaced. I will replace it to improve compatibility.

josephmancuso (approved):

PR #864, opened 2023-12-05
Add optional config path to ConnectionResolver and ConnectionFactory

### LINKED ISSUE #863: Allow for specifying a config path when using ConnectionResolver
I want to be able to specify a custom config path without setting an extra environment variable.

The current implementation of `ConnectionResolver` and `ConnectionFactory` does the following:

1. `ConnectionResolver` inits `ConnectionFactory`
2. When creating a new transaction, `ConnectionFactory` will run `make`, which calls `DB = load_config().DB`
3. `load_config()` will either use the default config path, or use the `DB_CONFIG_PATH` env var

**Additional context**
The load_config function already has a `config_path` argument. So, my proposal is to:
- Add an optional `config_path` argument to `ConnectionResolver`
- Add an optional `config_path` argument to `ConnectionFactory` 
- Pass through that `config_path` from `ConnectionResolver` to the initialization of `ConnectionFactory`
- Use the `config_path` in `load_config`

matt2930:
Closes #863

### DIFF
--- a/src/masoniteorm/connections/ConnectionFactory.py
+++ b/src/masoniteorm/connections/ConnectionFactory.py
@@ -4,9 +4,10 @@ from ..config import load_config
 class ConnectionFactory:
     """Class for controlling the registration and creation of connection types."""
 
-    _connections = {
-        #
-    }
+    _connections = {}
+
+    def __init__(self, config_path=None):
+        self.config_path = config_path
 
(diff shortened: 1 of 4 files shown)


josephmancuso (approved):

josephmancuso:
code review looks good

PR #865, opened 2023-12-09
Adds a create_or_first method to the Model class

JarriqTheTechie:
#### Added

- Adds `create_or_first ` 

#### Changed

- Modified `first_or_create ` to use `create_or_first` under the hood. The goal of this is to eliminate race conditions that can occur when using `first_or_create`. `first_or_create` has also been modified to first look for the first model matching the given criteria if none is found then it calls `create_or_first`. If a unique constraint violation occurs then it tries again to retrieve the model from the database. 

#### Context
Currently, `first_or_create` searches for a model matching the given criteria, if the model does not exist then it creates a new record. A problem can occur where a record is created matching the criteria during the time period between the database read and the database write. This would lead to a `QueryException` due to unique constraint fail. 

What `create_or_first` does is first attempts to create the model. If a unique constraint failure happens then it tries to retrieve the model from the database matching the values of the where clause. 

Both methods take matching arguments.

```python
create_or_first(wheres, creates)
first_or_create(wheres, creates):
```

This is useful in very busy/high concurrency environments where race conditions are more likely.

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -12,7 +12,7 @@ from inflection import tableize, underscore
 
 from ..collection import Collection
 from ..config import load_config
-from ..exceptions import ModelNotFound
+from ..exceptions import ModelNotFound, QueryException
 from ..observers import ObservesEvents
 from ..query import QueryBuilder
 from ..scopes import TimeStampsMixin
@@ -684,9 +684,9 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
         """
         return json.dumps(self.serialize(), default=str)
 
-    @classmethod
+classmethod
     def first_or_create(cls, wheres, creates: dict = None):
-        """Get the first record matching the attributes or create it.
+        """Attempts to find the first record matching the attributes or create it.
 
         Returns:
             Model
@@ -695,12 +695,31 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
             creates = {}
         self = cls()
         record = self.where(wheres).first()
+        if not record:
+            return self.create_or_first(wheres, creates)
+        return record
+
+    @classmethod
+    def create_or_first(cls, wheres, creates: dict = None):
+        """First attempts to create a record. If the record already exists then find the first record matching the
+        where clause.
+
+        Returns:
+            Model
+        """
+        if creates is None:
+            creates = {}
+        self = cls()
         total = {}
         total.update(creates)
         total.update(wheres)
-        if not record:
+
+        try:
             return self.create(total, id_key=cls.get_primary_key())
-        return record
+        except QueryException as e:
+            if "UNIQUE constraint failed" not in e.__str__():
+                raise e
+            return self.where(wheres).first()
 
     @classmethod
     def update_or_create(cls, wheres, updates):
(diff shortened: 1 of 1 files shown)


src/masoniteorm/models/Model.py
```diff
@@ -695,12 +695,31 @@ def first_or_create(cls, wheres, creates: dict = None):
             creates = {}
         self = cls()
         record = self.where(wheres).first()
+        if not record:
+            return self.create_or_first(wheres, creates)
```
josephmancuso, line 699: whats the reason for this change?

src/masoniteorm/models/Model.py
```diff
@@ -695,12 +695,31 @@ def first_or_create(cls, wheres, creates: dict = None):
...
+        return record
+
+    @classmethod
+    def create_or_first(cls, wheres, creates: dict = None):
```
josephmancuso, line 703: remind me what the difference is between this and first_or_create?

josephmancuso (changes requested):
its going to be way too confusing to have both a first or create method and a create or first method. before i decline it i need better reasoning here

src/masoniteorm/models/Model.py
```diff
@@ -695,12 +695,31 @@ def first_or_create(cls, wheres, creates: dict = None):
...
+        self = cls()
         total = {}
         total.update(creates)
         total.update(wheres)
-        if not record:
+
+        try:
             return self.create(total, id_key=cls.get_primary_key())
-        return record
+        except QueryException as e:
```
josephmancuso, line 719: i dont think we should be hijacking this exception its going to be thrown and overwrite legitimate  query exceptions right? like if an insert fails on the table?

also even if thats not the case i think you're trying to check for the existence of a specific database error is thrown which you cant do because all diff databases and versions throw diff errors

josephmancuso (changes requested):

JarriqTheTechie:
> its going to be way too confusing to have both a first or create method and a create or first method. before i decline it i need better reasoning here

Hey @josephmancuso i think the article I will link does a pretty quick explanation. 

https://laravel-news.com/firstorcreate-vs-createorfirst

Also the following YouTube video as well.
 https://youtu.be/uLXX48FKnuI?si=yJ11z2glTtYRnrNA

JarriqTheTechie:
> > its going to be way too confusing to have both a first or create method and a create or first method. before i decline it i need better reasoning here
> 
> Hey @josephmancuso i think the article I will link does a pretty quick explanation. 
> 
> https://laravel-news.com/firstorcreate-vs-createorfirst
> 
> Also the following YouTube video as well.
>  https://youtu.be/uLXX48FKnuI?si=yJ11z2glTtYRnrNA
> 
> 
> 

Further followup reference to this can be seen in rails active record where they have something very similar

find_or_create_by vs create_or_find_by

src/masoniteorm/models/Model.py
JarriqTheTechie: The solution to this would be to add a new exception. UniqueContraintViolationException to the connection classes. The different error codes for unique exception failures can be parsed on the connection class level. Allows for easier extensibility when adding support for new databases as well.

josephmancuso: i would be more comfortable with putting it in the connection class

josephmancuso:
ok makes sense then

josephmancuso:
going to close this. grew too stale for me. you can reopen but will also need to add tests

PR #866, opened 2023-12-09
Added only to Model

JarriqTheTechie:
#### Added 

- Adds only method to model class. This method allows you to get a subset of the model's attributes. 

#### Usage

```python
User.first_or_create(
      {"email": "abc@123.com"},
      {"email": "abc@123.com", "password": "password123"}
  ).only([
      "id",
      "email",
  ])
```

this would return the following

```python
{'id': 1, 'email': 'abc@123.com'}
```

In addition to the above usage, you can also use this as a way to remap data. We can alias attributes so that the results can match its destination. Example usage

```python
User.first_or_create(
      {"email": "abc@123.com"},
      {"email": "abc@123.com", "password": "password123"}
  ).only([
      "id as UserIdentifier",
      "email as EmailAddress",
  ])
``` 
this would return the following

```python
{'UserIdentifier': 2, 'EmailAddress': 'abc@123.com'}
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ setup(
     # Versions should comply with PEP440.  For a discussion on single-sourcing
     # the version across setup.py and the project code, see
     # https://packaging.python.org/en/latest/single_source_version.html
-    version="2.19.2",
+    version="2.20.0",
     package_dir={"": "src"},
     description="The Official Masonite ORM",
     long_description=long_description,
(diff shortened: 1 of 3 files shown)


josephmancuso:
made a small change here if you just want to pull 1 column by not needing to pass a list

```python
User.first_or_create(
      {"email": "abc@123.com"},
      {"email": "abc@123.com", "password": "password123"}
  ).only("email")
```

josephmancuso:
and i added tests

josephmancuso (approved):

PR #869, opened 2024-02-12
BaseRelationship: s/self/instance

pmn4:
when migrating from Orator to Masonite, several of my relationship methods were not working. Stepping through the code, I found that when referencing the `class` via `self.__class__` inside the `User` class, `self` was an instance of `BelongsTo`, not `User`. This PR aims to pass an instance of `User` instead.

``` Python
class User(Model):

    @belongs_to
    def parent(self):
        # ⚠️ self is an instance of BelongsTo ⚠️ 
        # expecting self to be an instance of User
        return self.__class__
```

### DIFF
--- a/src/masoniteorm/relationships/BaseRelationship.py
+++ b/src/masoniteorm/relationships/BaseRelationship.py
@@ -52,7 +52,7 @@ class BaseRelationship:
             object -- Either returns a builder or a hydrated model.
         """
         attribute = self.fn.__name__
-        relationship = self.fn(self)()
+        relationship = self.fn(instance)()
         self.set_keys(instance, attribute)
         self._related_builder = relationship.builder
 
--- a/tests/mssql/builder/test_mssql_query_builder_relationships.py
+++ b/tests/mssql/builder/test_mssql_query_builder_relationships.py
@@ -40,6 +40,14 @@ class User(Model):
     def profile(self):
         return Profile
 
+    @belongs_to("id", "parent_dynamic_id")
+    def parent_dynamic(self):
+        return self.__class__
+
+    @belongs_to("id", "parent_specified_id")
+    def parent_specified(self):
+        return User
+
 
 class BaseTestQueryRelationships(unittest.TestCase):
     maxDiff = None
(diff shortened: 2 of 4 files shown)


josephmancuso:
can you expand on what this is fixing?

pmn4:
oops, this isn't a good example. the user method should return the User class. I tried to make my example simpler, and instead made it confusing!
instead, consider a User class that has a parent relationship, where parent is another record in the User table.
you could say:
``` Python
@belong_to 
def parent(self):
    return User
```

or

``` Python
@belong_to 
def parent(self):
    return self.__class__
```

src/masoniteorm/relationships/BaseRelationship.py
```diff
@@ -52,7 +52,7 @@ def __get__(self, instance, owner):
             object -- Either returns a builder or a hydrated model.
         """
         attribute = self.fn.__name__
-        relationship = self.fn(self)()
+        relationship = self.fn(instance)()
```
josephmancuso, line 55: should this be the intance variable or the owner variable? i think owner. instance would be the class but owner would be the initialized object.

josephmancuso (changes requested):

tests/mssql/builder/test_mssql_query_builder_relationships.py
```diff
@@ -40,6 +40,14 @@ def articles(self):
     def profile(self):
         return Profile
 
+    @belongs_to("id", "parent_dynamic_id")
+    def parent_dynamic(self):
+        return self.__class__
```
pmn4, line 45: previously, this test would have failed

tests/mssql/builder/test_mssql_query_builder_relationships.py
```diff
@@ -64,6 +72,32 @@ def test_has(self):
...
+    def test_has_reference_to_self_using_class(self):
+        builder = self.get_builder()
+        sql = builder.has("parent_specified").to_sql()
+
+        print(sql)
+
+        self.assertEqual(
+            sql,
+            """SELECT * FROM [users] WHERE EXISTS ("""
+            """SELECT * FROM [users] WHERE [users].[parent_specified_id] = [users].[id]"""
```
pmn4, line 91: any issue here, not having an alias for `[users]` when we self-join?

tests/mysql/model/test_model.py
```diff
@@ -316,7 +316,7 @@ def test_attribute_check_with_hasattr(self):
         self.assertFalse(hasattr(Profile(), "__password__"))
 
 
-if os.getenv("RUN_MYSQL_DATABASE", None).lower() == "true":
+if os.getenv("RUN_MYSQL_DATABASE", "false").lower() == "true":
```
pmn4, line 319: _(my testing environment wasn't set up, so this caused an error, saying that `lower` is not a function on `NoneType`)_

src/masoniteorm/relationships/BaseRelationship.py
pmn4: `owner` appears to be the class, `User` in my example, whereas `instance` is an instance of `owner`. `self.fn` is a pointer to an instance method (Ruby vernacular) or prototype function (Node) -- apologies, I don't know the corresponding term in Python -- so I _believe_ `instance` is correct

josephmancuso (approved):
I think this is fine but i need more time to test this change

PR #879, opened 2024-04-02
wip: hasmany performance fix

josephmancuso:


josephmancuso (commented):
we need to modify this on all relationship classes right?

PR #880, opened 2024-04-03
Fix/relationship performance joe attempt

josephmancuso:


src/masoniteorm/relationships/HasMany.py
```diff
@@ -28,5 +28,5 @@ def set_keys(self, owner, attribute):
 
     def register_related(self, key, model, collection):
         model.add_relation(
-            {key: collection.where(self.foreign_key, getattr(model, self.local_key))}
+            {key: collection.get(getattr(model, self.local_key)) or Collection()}
```
josephmancuso, line 31: hmm. is this right? before it was fetching by the foreign_key and now by the local_key?

when working with test records and all the records have ids 1-4 it can be difficult to make sure we're fetching the right related records. have to test this

tests/sqlite/builder/test_sqlite_query_builder_eager_loading.py
```diff
@@ -96,4 +96,5 @@ def test_with_multiple_per_same_relation(self):
         builder = self.get_builder()
         result = User.with_("articles", "articles.logo").where("id", 1).first()
         self.assertTrue(result.serialize()["articles"])
+        print("pppppp", result.serialize()["articles"][0])
```
josephmancuso: remove

PR #881, opened 2024-04-05
Add some type-hints to Model.py

JarriqTheTechie:


JarriqTheTechie:
When adding multiple return types I will use Union instead of pipe operator.

src/masoniteorm/models/Model.py
```diff
@@ -272,7 +272,7 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
 
     __cast_map__ = {}
 
-    __internal_cast_map__ = {
+    __internal_cast_map__: list[str, Any] = {
```
josephmancuso: is this a list or is it a dict?

josephmancuso (changes requested):

josephmancuso:
also tests are failing

JarriqTheTechie:
> also tests are failing

Tests are failing due to issues with backwards compatibility with older versions of python. That's why I'll use Union types instead of | as a type separator.

josephmancuso:
please fix tests failing. i'd like to merge this in

josephmancuso:
closing bc of tests failing

PR #882, opened 2024-04-10
feat: implement json search

eaguad1337:


src/masoniteorm/query/grammars/SQLiteGrammar.py
```diff
@@ -24,7 +24,7 @@ class SQLiteGrammar(BaseGrammar):
     }
 
     column_strings = {
-        "select": '{table}."{column}"{alias}{separator}',
+        "select": "{table}.`{column}`{json_path}{alias}{separator}",
```
josephmancuso, line 27: why are are these " replaced with a ` symbol.

sqlite should be "" i think

tests/sqlite/grammar/test_sqlite_select_grammar.py
```diff
@@ -461,3 +461,13 @@ def or_where_null(self):
 
     def select_distinct(self):
         return """SELECT DISTINCT "users"."group" FROM "users\""""
+
+    def test_can_compile_json_where(self):
+        to_sql = self.builder.where("options->age", 18).to_sql()
+        self.assertEqual(to_sql, "SELECT * FROM \"users\" WHERE \"users\".`options`->'$.age' = '18'")
```
josephmancuso, line 467: this syntax is invalid and i cant get this to run on sqlite

josephmancuso (changes requested):

tests/sqlite/grammar/test_sqlite_select_grammar.py
eaguad1337: Weird. I swear I tested it.

PR #883, opened 2024-05-02
Add configuration option to enable or disable foreign keys check for all 4 types of database connections

martinszemitis:
**Problem:**
When defining migrations for SQLite and, for example, setting cascading on delete it dosn't work as foreign keys are not enabled. According to documentations https://www.sqlite.org/foreignkeys.html they should be enabled manually.

**Solution:**
To not rely on changes how DBMS handles that now and in the future, developer not can enable / disable foreign keys in database connection explicitly. That now would be possible for all 4 types of database.


josephmancuso (approved):
LGTM

PR #892, opened 2024-09-10
Tests query=True should return query builder

circulon:
This is mainly cleanup and aligning standard return values for test cleanliness


josephmancuso (approved):

josephmancuso:
fixed conflicts

josephmancuso (approved):

PR #893, opened 2024-09-11
Fix find cannot use scopes

circulon:
This PR fixes the issue when using a mixin scope like SofDelete with .find() will not find the record.
This is due to only the global scope being applied regardless.

This impacts records that have defaults that the gloal scop excludes
example:
- create a record  that sets active=False by default
- the global scope is active=True
- Record.find(some_id) # works as expected with glpba; scope and finds no record
- Record.is_inactive().find(some_id)  # should wprk by setting scope where active=False  to allow finding the record but is not applied


circulon:
@josephmancuso @eaguad1337 
I think this is completed now
cheers

josephmancuso (approved):

PR #896, opened 2024-10-18
Added SSL configuration in PostgresConnection

### LINKED ISSUE #875: Postgres psycopg2 TLS connectivity

I don't think connecting to a Postgres Database over TLS is currently possible based on what code I saw when connecting via psycopg2. I would like to use `sslmode, sslrootcert options and that doesn't appear to be possible. 

Please confirm if this is a valid feature to add, if so I can submit some code to enable this ability. I believe all we need to do is add on a general **kwargs parameter to the make_connection function and pass that along since psycopg2 can handle it. 

**Additional context**

Great project by the way! Really appreciate the work put into it!

- [ ] Is this a breaking change? No

josephmancuso:
Closes #875 

```python
"postgres": {
    "host": "127.0.0.1",
    "driver": "postgres",
    "database": "masonite",
    "user": "root",
    "password": "",
    "port": 5432,
    "log_queries": False,
    "options": {
      "sslmode": "...",
      "sslcert": "...",
      "sslkey": "..."
    }
```

### DIFF
--- a/src/masoniteorm/connections/PostgresConnection.py
+++ b/src/masoniteorm/connections/PostgresConnection.py
@@ -65,7 +65,7 @@ class PostgresConnection(BaseConnection):
             host=self.host,
             port=self.port,
             options=f"-c search_path={schema}" if schema else None,
-            sslmode=self.options.get("sslmode", "prefer"),
+            sslmode=self.options.get("sslmode"),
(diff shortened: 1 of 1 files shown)


josephmancuso:
needs manual testing

josephmancuso:
impolemented in another PR

PR #900, opened 2024-10-22
Add connection pooling to MySQL and Postgres

josephmancuso:
- [x] mysql
- [x] Postgres

with sqlite i saw a net decrease in the performance


josephmancuso:
mysql is done

PR #901, opened 2024-10-25
fixed .on_null and .on_not_null had to be last criteria

circulon:
This PR fixes an issue where a join clause like this:
```python
user_join = (
            JoinClause(f"{u_table} as ut")
            .on("user_id", "=", "ut.id")
            .on_value("ut.email", "=", email)
            .on_null("ut.other_id")
            .on_value("ut.active", "=", True)
        )
```
produces a query similar to this:
```sql
SELECT * FROM "my_model" 
    INNER JOIN "user" AS "ut" 
        ON "my_model"."user_id" = "ut"."id" 
        AND "ut"."email" = 'email@fdomain.com' 
        AND "ut"."other_id" IS NULLAND "ut"."active" = 'True' 
```
Notice there is no space between the "IS NULL" and the next "AND" criterua.
This means that currently the .on_null() and .on_not_null() clauses must always be the last criteria
and as such there cannot be more than 1 of either in the join clause.

### DIFF
--- a/src/masoniteorm/query/grammars/BaseGrammar.py
+++ b/src/masoniteorm/query/grammars/BaseGrammar.py
@@ -251,13 +251,13 @@ class BaseGrammar:
                         on_string += f"{keyword} {self._table_column_string(clause.column1)} {clause.equality} {self._table_column_string(clause.column2)} "
                     else:
                         if clause.value_type == "NULL":
-                            sql_string = self.where_null_string()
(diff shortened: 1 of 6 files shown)


circulon:
@josephmancuso @eaguad1337 
I think this is ready to go now

josephmancuso (approved):

josephmancuso:
LGTM

PR #903, opened 2024-10-29
Fix has many through relationship not working

circulon:
This PR issues with the HasManyThrough relationship to not return related models

Issues addressed:
- eager loading did not work
- related loading would only sometimes work
- where_xxx of relation built incorrect query criteria

Other things
- use vars for table names and beter readability
- added some more metod docstrings
- added tests to prove eager/related loading is working as expected


src/masoniteorm/relationships/HasManyThrough.py
```diff
@@ -57,33 +57,46 @@ def __get__(self, instance, owner):
             if attribute in instance._relationships:
                 return instance._relationships[attribute]
 
-            result = self.apply_query(
+            result = self.apply_related_query(
```
josephmancuso, line 60: why is this renamed?

josephmancuso (changes requested):

src/masoniteorm/relationships/HasManyThrough.py
```diff
@@ -57,33 +57,46 @@ def __get__(self, instance, owner):
...
-        # select * from `countries` inner join `ports` on `ports`.`country_id` = `countries`.`country_id` where `ports`.`port_id` is null and `countries`.`deleted_at` is null and `ports`.`deleted_at` is null
-        result = distant_builder.join(
-            f"{self.intermediary_builder.get_table_name()}",
-            f"{self.intermediary_builder.get_table_name()}.{self.foreign_key}",
-            "=",
-            f"{distant_builder.get_table_name()}.{self.other_owner_key}",
-        ).where(f"{self.intermediary_builder.get_table_name()}.{self.local_owner_key}", getattr(owner, self.other_owner_key)).get()
 
-        return result
+        dist_table = distant_builder.get_table_name()
```
josephmancuso: lets use full named variables here. distant_table, intermediate_table.

1. shorthand variable names drives me up a wall
2. the common developers attempt to shorthand variable names in an attempt to make code .. ?( more readable?) does the opposite of what it tries to do

src/masoniteorm/relationships/HasManyThrough.py
circulon: @josephmancuso 

this was renamed as the BaseRelationship methos has a completely different sigbature and parameter requirements
from BaseRelationship on line 72
```python
def apply_query(self, foreign, owner, foreign_key, local_key):
```
Though the usage of this method is similar the parameter types are entirely different and the number of required parameters is different too.
Renaming the method means it will not be used as a subclass overide in future.

src/masoniteorm/relationships/HasManyThrough.py
```diff
@@ -104,51 +117,144 @@ def make_builder(self, eagers=None):
...
+    def attach_related(self, current_model, related_record):
+        raise NotImplementedError(
+            "HasOneThrough relationship does not implement the attach_related method"
+        )
 
-        if not builder._columns:
-            builder = builder.select("*")
+    def query_has(self, current_builder, method="where_exists"):
+        dist_table = self.distant_builder.get_table_name()
+        int_table = self.intermediary_builder.get_table_name()
```
josephmancuso: same here

src/masoniteorm/relationships/HasManyThrough.py
```diff
@@ -104,51 +117,144 @@ def make_builder(self, eagers=None):
...
-                Collection(relation._get_value(self.local_key)).unique(),
+            return self.distant_builder.where_in(
+                f"{int_table}.{self.local_key}",
+                Collection(relation._get_value(self.local_owner_key)).unique(),
             ).get()
         else:
-            return builder.where(
-                f"{builder.get_table_name()}.{self.foreign_key}",
+            return self.distant_builder.where(
+                f"{int_table}.{self.local_key}",
```
josephmancuso: same here

src/masoniteorm/relationships/HasManyThrough.py
```diff
@@ -104,51 +117,144 @@ def make_builder(self, eagers=None):
...
+
+    def query_where_exists(self, current_builder, callback, method="where_exists"):
+        dist_table = self.distant_builder.get_table_name()
+        int_table = self.intermediary_builder.get_table_name()
+
+        getattr(current_builder, method)(
+            self.distant_builder.join(
+                f"{int_table}",
+                f"{int_table}.{self.foreign_key}",
+                "=",
```
josephmancuso, line 217: same here

src/masoniteorm/relationships/HasManyThrough.py
```diff
@@ -104,51 +117,144 @@ def make_builder(self, eagers=None):
...
             f"{self.attribute}_count",
             lambda q: (
                 (
                     q.count("*")
                     .join(
-                        f"{self.intermediary_builder.get_table_name()}",
-                        f"{self.intermediary_builder.get_table_name()}.{self.foreign_key}",
+                        f"{int_table}",
+                        f"{int_table}.{self.foreign_key}",
                         "=",
```
josephmancuso, line 242: same here

src/masoniteorm/relationships/HasManyThrough.py
```diff
@@ -157,47 +263,3 @@ def get_with_count_query(self, builder, callback):
...
-            query.join(
-                f"{self.intermediary_builder.get_table_name()}",
-                f"{self.intermediary_builder.get_table_name()}.{self.foreign_key}",
-                "=",
-                f"{query.get_table_name()}.{self.other_owner_key}",
-            ).where_column(
-                f"{current_query_builder.get_table_name()}.{self.local_owner_key}",
-                f"{self.intermediary_builder.get_table_name()}.{self.local_key}",
-            )
-        ).when(callback, lambda q: (callback(q)))
```
josephmancuso, line 203: why are all these deleted?

josephmancuso (changes requested):

src/masoniteorm/relationships/HasManyThrough.py
circulon: yep np

src/masoniteorm/relationships/HasManyThrough.py
circulon: They are not deleted.
These lines are now at 221 - 225 respectively
```python
                f"{intermediate_table}.{self.local_key}",
                f"{current_builder.get_table_name()}.{self.local_owner_key}",
            )
            .when(callback, lambda q: (callback(q)))
        )
```
Lines  200 & 201 had readability changes.
Line 203 moved the `.when()` from the current query builder to the distant builder as thats where the criteria should be

src/masoniteorm/relationships/HasManyThrough.py
circulon: > lets use full named variables here. distant_table, intermediate_table.

Yes no worries
 
> 1. shorthand variable names drives me up a wall

Yeah I can understand that  ;)

> 2. the common developers attempt to shorthand variable names in an attempt to make code .. ?( more readable?) does the opposite of what it tries to do

True, if its done "too short" it can reduce understanding and readability more than the original more verbose version.

All fixed now

src/masoniteorm/relationships/HasManyThrough.py
circulon: @josephmancuso 
Based on this I had a look at all the other relationships and all of the (except HasOneThrough)
have method signatures for `apply_query) that do not match the BaseRelationship method signature.

Just thinking uot loud here... 
We should probably look at aligning these classes with the BaseRelationship in a separate PR

circulon:
@josephmancuso 
Is there anthing else i/we need to look at/discuss to get this merged please?

josephmancuso (approved):

Issue #902, opened 2024-10-26
Model queryBuilder should only select columns for that model


circulon:
**Describe the bug**
By default the QueryBuilder attached to a model via `Model.builder.xxx 
uses `SELECT * FROM <model_table> ...` as the default select criteria.

This is an issue when adding join clauses (and possibly other clauses) either via scopes or as part of a models query.
This causes the Model to be loaded with every column returned by the query,

Additionally any column names from other tables that have the same name as the ones in the model (id is a prime example)
will be updated with the value from the other tables column creating a model that is not the same as the row in the DB. 

**To Reproduce**
Assume the following models:
```python
class Company(Model):
    id: int
    name: str
    email: string

class User(Model):
    id: int
    company_id: int
    first_name: string
    last_name: string
    email: string

    @scope
    def with_email(self, query, email):
        c_table = Company.get_table_name()
        company_join = (
            JoinClause(f"{c_table} as ct")
            .on("company_id", "=", "ct.id")
            .on_value("user.email", "=", email)
        ) 
        return query.join(company_join)
```
assume the following tables and rows:

Table: company
| id | name | email |
| ---- | ---------- | -------------------- |
| 1 | Acme | support@acme.com |
| 2 | Ego | admin@ego.net |
| 3 | Tools R Us | info@toolsrus.org |

Table: user
| id | company_id | first_name | last_name | email |
| ---- | ---- | ---------- | ---------- | -------------------- |
| 1 | 2 | Bob | Down | bob@ego.net |
| 2 | 2 | James | Brown | james@ego.net |
| 3 | 1 | Steve | Jones | steve@acme.com |
| 4 | 3 | Mark | Black | mark@toolsrus.org |
| 5 | 3 | James | Jameson | james@toolsrus.org |

query to find "James" from "Tools R Us"
```python
query = User.with_email("james@toolsrus.org")
print(query.to_sql())
user = query.first()
print(user.serialize())
```
Generated query will be similar to:
```sql
SELECT * FROM "user" INNER JOIN "company" AS "ct" ON "user"."company_id" = "ct"."id" AND "user"."email" = 'james@toolsrus.org'
```
Row data returned looks similar to this:
| id | company_id | first_name | last_name | email | id | name | email |
| ---- | ---- | ---------- | ---------- | -------------------- | ---- | ---------- | -------------------- |
| 5 | 3 | James | Jameson | james@toolsrus.org | 3 | Tools R Us | info@toolsrus.org |

[...]

PR #904, opened 2024-10-30
Added ability so specify default select criteria for Models

circulon:
Based on discussion in #902 

Specify default select criteriaa for the Model by
overriding  get_selects() in the subclass

@josephmancuso 
do we need more tests than this?
ie for all grammars?


josephmancuso (approved):

josephmancuso:
no this looks good

circulon:
@josephmancuso 
I will have alook at the docs regarding this soon

PR #907, opened 2024-11-11
fix nested relation

pbellerive:


josephmancuso:
im going to work on moving this mapping into. the relation classes

PR #912, opened 2024-12-01
allow override of model default selects

circulon:
This PR addresses generating queries with cusom select criteria or aggregation while also defining default select criteria in the model

This moves the addition of the default select columns from the creating the initial builder instance in the model  to the querybuilder where checking if the defaults are actually required is done instead. 

Additional Cleanup:
- Removed some unused imports  from tests
- Removed some unused commented out parameters from tests
- Sorted imports on some tests

### DIFF
--- a/src/masoniteorm/models/Model.py
+++ b/src/masoniteorm/models/Model.py
@@ -354,7 +354,7 @@ class Model(TimeStampsMixin, ObservesEvents, metaclass=ModelMeta):
             dry=self.__dry__,
         )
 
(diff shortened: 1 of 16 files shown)


circulon:
@josephmancuso 
Can we get this reviewed and merged please?
This is currently affecting aggregate functions usage in our test code.

josephmancuso:
Yes I'll put this in the schedule

josephmancuso (approved):

Issue #925, opened 2025-02-22
QueryBuilder .find() throws Attribute Error when not used with a model


circulon:
**Describe the bug**
When using a plain QueryBuilder the `.find()`, `.find_or_fail()` and `.find_or_404()` will throw an `AttributeError`
the error is `AttributeError: 'NoneType' object has no attribute 'get_primary_key'`
 

**To Reproduce**
```python
builder = QueryBuilder()
try:
    record = builder.find(100)
    record = builder.find_or_fail(100)
    record = builder.find_or_404(100)
except ModelNotFound as e:
    print("model not found")
``

**Expected behavior**
when using a plainQueryBuilder without a model,  the methods should either:
return the found recod 
or 
 raise a `ModelNotFound` exception if the record is not found 

**Screenshots or code snippets**.
NA

**Desktop (please complete the following information):**
 - OS: MacOS
 - Version 12.7

**What database are you using?**
 - Type:  all
 - Version NA
 - Masonite ORM 2.24.0

**Additional context**
NA

PR #926, opened 2025-02-23
Fix 925

circulon:
This PR addresses #925 

Iy also fixes using .findxxx methods with a list (or tuple) of values as per the documentation.

As a bonus we can now use `.find(column="xx")` as a shorcut for a basic equality in `.wher()` and `.where_in()`  

example:
```python
user = User.find(10, column="age")
# is equivalent to
user = User.where("age", 10).first()

users = User.find([10, 20, 30], column="age")
# is equivalent to
users = User.where_in("age", [10, 20, 30]).get()
```
Documentation updates will be done if this PR is accepted noting the above shortcuts


circulon:
@josephmancuso  @eaguad1337 
Can we get this reviewed please?

josephmancuso (approved):

Issue #927, opened 2025-02-23
Cannot attach unsaved model


circulon:
**Describe the bug**
Currently if you try to `.attach()` a model that has not yet been saved an error occurs.
This is specifically related to having foreign keys that are NOT nullable.

**To Reproduce**
with these models
```python

class Profile(Model):
    __table__ = "profile"

   # user_id is a non nullable integer

    @belongs_to("id", "user_id")
    def user(self):
        return User
   
class User(Model):
    __table__ = "user"

    @has_one("user_id", "id")
    def profile(self):
        return Profile
```
attaching an unsaved record fails as the "user_id" is NULL
```python
profile = Profile.fill({
    "role": "admin",
})
user = User.create({
    "name": "fred",
})
user.attach("profile", profile)
# throws a query exception "user_id" cannot be NULL
```

**Expected behavior**
Unsaved records should have the required keys linking them set during attach()

**Screenshots or code snippets**
NA

**Desktop (please complete the following information):**
 - OS:  MacOS
 - Version 12.7

**What database are you using?**
 - Type: All
 - Version NA
 - Masonite ORM 2.24.0

**Additional context**
NA

PR #928, opened 2025-02-23
Fix 927

circulon:
Addresses #927 

Moved the attach logic to the Relationships
Populat the correct key/s when creating the model 

Cleaned up where exceptions for required methods are generated from


circulon:
@josephmancuso  @eaguad1337 
Can we get this reviewed please?

josephmancuso (approved):

Issue #935, opened 2025-03-27
use pre-commit for automaic code checks


circulon:
**Describe the feature as you'd like to see it**
Currently the only time the 'black' and 'isort' flows are used to tidy code to be committed is if its done manually by the dev prior to the commit bein done.
Linting via flake8 also has to be done manually but is included in the github checks when the branch is pushed

If these manual tasks (like make format) are not done it can  leads to things like (but not limited to):
- unused imports left in 
- unused variables left in
- long strings ie imports or chained expressions which can be hard to read/parse
- imports not grouped or sorted 

**What do we currently have to do now?** 
A better alternative is to use the `pre-commit` package whic can run all these tools over the staged code and inplement tidying and cleanup via `black` and `isort` 
Additionally `flake8` will lint your code and notify of issues which must be addressed before the commit is accepted

The `interrogate` package is also useful when combined with `pre-commit` as is forces cpmmenting of classes amd methods to be done.  

We use `pre-commit` on all our projects and find it to be an easy way to integrate tools to keep our code clean and readable and organised.

**Additional context**
@josephmancuso Im happy to put a PR together which basic setup for `pre-commit` with the black, isort and flake8` modules

- [ ] Is this a breaking change?
No

josephmancuso:
Sounds great

circulon:
> Sounds great

ok in which case as `3.0` has been tagged should I be branching from `3.x` even though its not uptodate with `2.0`? 
ie the 3.0 tag is based off  2.18.1 which is fairly old now.

or should I do separate PR for 2.0 and 3.x?

PR #939, opened 2025-03-29
Transition 2.0 to 3.0

circulon:
Got the right Destination branch this time..

This PR includes all the features and fixes from the 2.0 branch
Also includes features from the current 3.x branch.

Additionally includes pre-commit setup as discussed in #935 
Linted and formatted the entire codebase as necessary.

@josephmancuso


circulon:
@josephmancuso 
I think this is ready now ;)

Been testing it without issues ;)

Will have to update the docs with steps for upgrade
Luckily there are only a couple of things to address

circulon:
@josephmancuso  LGTM please

src/masoniteorm/__init__.py
```diff
@@ -1,2 +1 @@
 from .models import Model
-from .factories.Factory import Factory
```
josephmancuso, line 2: why is this removed?

src/masoniteorm/commands/MigrateRefreshCommand.py
```diff
@@ -33,11 +31,11 @@ def handle(self):
         if self.option("seed") == "null":
             self.call(
                 "seed:run",
-                f"None --directory {self.option('seed-directory')} --connection {self.option('connection', 'default')}",
```
josephmancuso, line 36: Is this correct? Shouldn't this default to the `default` connection?

src/masoniteorm/config.py
```diff
@@ -12,8 +12,11 @@ def load_config(config_path=None):
         2. else try to load from default config_path: config/database
     """
     selected_config_path = (
-        config_path or os.getenv("DB_CONFIG_PATH", None) or "config/database"
+        os.getenv("DB_CONFIG_PATH", None) or config_path or "config/database"
```
josephmancuso: Change this to 

```python
os.getenv("DB_CONFIG_PATH", config_path) or "config/database"
```

josephmancuso (changes requested):

src/masoniteorm/commands/MigrateRefreshCommand.py
josephmancuso: right now this would default to the None connection right?

src/masoniteorm/config.py
josephmancuso: its simpler like that

src/masoniteorm/__init__.py
circulon: The `Factory` class is only rewuired for Seeding as far as I can tell.

From a production environment perspective this is redundant
I added a [seeder] to the setup which will provide the Seeding capability if required.

I also noted that the Factory class should be imported from its package in the CHANGELOG in the  breaking changes section

src/masoniteorm/commands/MigrateRefreshCommand.py
circulon: @josephmancuso 
unsure about this?
you altered it in commit 18de4662

circulon: @josephmancuso 
seems to work fine from the command line for me 

Do you need to do some testing?
Or can we resolve this?

circulon:
@josephmancuso  LGTM

circulon:
@josephmancuso 

Can we move this forward please?

circulon:
@josephmancuso 
Is there anthing else that needs addressing to GTM?

josephmancuso (approved):

PR #950, opened 2025-12-06
Fix 949

### LINKED ISSUE #949: increments() and *_increments() forces column as primay key
**Describe the bug**
Currently trying to add a `increments()` or `big_increments()` automatically sets that column as the primary key.
This creates an issue when using any other data type as a prinmary key (eg uuid) and also using a
big_increments column in the same table.
This scenario will raise a `masoniteorm.exceptions.QueryException: table "my_table" has more than one primary key`  

Ideally a tables primary key should never be assumed and `.big_increments()` should used with `.primary()` as is the case with all other datatypes that can be used as a primary key.

**To Reproduce**
example Migration code throws 
masoniteorm.exceptions.QueryException: table "invoice" has more than one primary key
```python
        with self.schema.create("invoice") as table:
            table.uuid("id").primary()
            table.big_increments("reference")
            table.uuid("company_id")
            table.string("file_location")
            table.timestamps()
```
A workaround thst works in postgres.
NOTE: this does not work with sqlite as the `big_increments` does not set it as an auto-increment column
```python
        with self.schema.create("invoice") as table:
# note the id column is not primary
            table.uuid("id")
            table.big_increments("reference")
            table.uuid("company_id")
            table.string("file_location")
            table.timestamps()

# drop the primay key index and assign the id column instead 
        with self.schema.table("invoice") as table:
            table.drop_primary("invoice_reference_primary")
            table.primary("id")
```

**Expected behavior**
big_increments() columns should be usable as a primary key or not

N/A

 - OS: MacOs
 - Version 12.7

 - Type: Postgres, MySQL, SQLite
 - Version any
 - Masonite ORM: latest 3.x branch

**Additional context**
N/A

circulon:
This fixes the issue where using an *increments() colum automatically assigned it as the primary key.
This aligns the `*increments()` with all other datatypes that can be used as a primary key or not.

There are several datatype tweaks for types that are not supported on that platform.
Notably on Sqlite AUTOINCREMENT is only supported for primary keys. So an exception is raised if `.increments()` is used on a column that is not markerd with `.primary()`

I added tests for creating the `migrations` as this is a common usage point for `.increments()` accross all platforms  

Note: this is a breaking change for those developers using *incremenrs() as a primary key.
for this case just adding .primary() to the column definition will duplocate current behaviour. 

closes #949


circulon:
@josephmancuso @eaguad1337 

Can you review this please or is there anything else we need to cover with this?

src/masoniteorm/schema/Schema.py
```diff
@@ -87,8 +88,14 @@ def on(self, connection_key):
         Returns:
             cls
         """
-        resolver = load_config(config_path=self.config_path).DB
-        self.connection_details = resolver.get_connection_details()
+        if not self.connection_details:
```
eaguad1337: This should be: If connection_details is not provided, load from config. If provided, use it directly.

CHANGELOG.md
```diff
@@ -9,15 +9,20 @@
 
 #### Breaking Changes
 
+- Using `*increments()` for primary key definition now requires adding `.primay()` to the column definition.
```
eaguad1337: this should be primary()

CHANGELOG.md
```diff
@@ -9,15 +9,20 @@
...
 - Seeding depencies are now in a separate`[seeder]` extension
 - `Factory` class must now be imported from the sub-package `masoniteorm.factories`
   
 ### Fixed
 
- - Model `update` and `delete` not casting passed values
+- QueryBuilder does not require `database.py` when passing connection detain in directly
+- Import DB (ConnectionResolver) from config even if inline connection details are provided   
+- `*increments()` can be used fir non primary key columns if supported by the platform
+- `*increments()` can be used fir non primary key columns if supported by the platform
```
eaguad1337: lines 22 y 23 are duplicated

eaguad1337: also, fir -> for?

eaguad1337 (changes requested):

CHANGELOG.md
circulon: Fixed

CHANGELOG.md
circulon: Fixed

circulon:
Hey @eaguad1337 
can we get this reviewed please?

Additionally once this goes through can you update the 3.0.0 tag to point to the current 3.x branch HEAD as the current tag just doesnt work as expected.

Cheers

